PR #1, opened 2019-02-20
Add VBase Client to save the start time of an A/B Test

tergol1:
To get properly data from Storedash the it's necessary to specify the start time of an A/B Test. So is necessary to store this information.

For the purpose specified, the problem can be solved by saving the data in the VBase.

* Start an A/B Test
* Initialize the analysis of the Test through the route for `initializeAbTest`
* Try get the partial result of the test through the route `abTestStatus`

#### Types of changes
* [ ]  Bug fix (non-breaking change which fixes an issue)
* [x]  New feature (non-breaking change which adds functionality)
* [ ]  Breaking change (fix or feature that would cause existing functionality to change)
* [ ]  Requires change to documentation, which has been updated accordingly.


node/index.ts
```diff
@@ -1,5 +1,5 @@
 import {ColossusContext} from 'colossus'
-import {ABtest} from '../TestCase'
+import { initializeABtest, ABTestStatus} from '../TestCase'
```
pedro823: _nit_: Linting. If you're using VSCode, pressing `option+shift+F` formats the whole file so you don't need to worry about these.

ab-test/date-formater.ts
```diff
@@ -0,0 +1,11 @@
...
+    var d = new Date(date),
+        month = '' + (d.getMonth() + 1),
+        day = '' + d.getDate(),
+        year = d.getFullYear()
+
+    if (month.length < 2) month = '0' + month
+    if (day.length < 2) day = '0' + day
+
+    return [year, month, day].join('-');
+}
```
pedro823: [Javascript's got you covered in this whole function](https://stackoverflow.com/questions/3552461/how-to-format-a-javascript-date#answer-34015511)

TestCase.ts
```diff
@@ -1,12 +1,34 @@
...
 
+    await vbase.save('ABTest', 'currentTestAB.json', {
+        name: testId,
+        timeStart: beginning
+    })
+}
+
+export async function ABTestStatus(ctx: ColossusContext)
+{
+    const vbase = await new VBase(ctx.vtex)
```
pedro823: This doesn't belong here. This function shouldn't create a new object every time it is called. Instead, this function should deconstruct `ctx` to get this vbase object, like so:
```ts
const { resources: { vbase } } = ctx
```
This object `resources` should be initialized on the index.ts, and declarated somewhere else, such as `resources/index.ts`.

Refer to [support-browser-session's code](https://github.com/vtex/support-browser-session/blob/master/node/index.ts): The Resources class is created when the request comes in. This initializes all the resources i'll need at the same time. The resource initialization is [here](https://github.com/vtex/support-browser-session/blob/master/node/resources/index.ts), and it uses the vbase object [here](https://github.com/vtex/support-browser-session/blob/master/node/utils/oneTimeToken.ts).

node/index.ts
```diff
@@ -8,11 +8,19 @@ export default {
     }
   },
   routes: {
-    abTest: async (ctx: ColossusContext) => {
+    initializeAbTest: async (ctx: ColossusContext) => {
       ctx.set('Cache-Control', 'no-cache')
       ctx.response.status = 200
```
pedro823: Two things:

1. You can set this with `ctx.status = 200` instead.
2. You're setting the 200 OK header without even doing the computation, how do you know it is going to succeed? 😮

node/index.ts
```diff
@@ -8,11 +8,19 @@ export default {
...
 
-      var ResultAB = await ABtest(ctx)
+      await initializeABtest(ctx)
+
+      ctx.response.body = 'A/B Test initialized'
```
pedro823: you can use `ctx.body` for that.

node/index.ts
```diff
@@ -8,11 +8,19 @@ export default {
     }
   },
   routes: {
-    abTest: async (ctx: ColossusContext) => {
+    initializeAbTest: async (ctx: ColossusContext) => {
```
pedro823: _nit_: `index.ts` should be a place where you bundle all the things you need, not the place to declare complex functions. Instead, create a file called `initializeAbTest.ts` where you declare this function, and then import it here on `index.ts`.

ab-test/evaluate.ts
```diff
@@ -30,7 +30,7 @@ export async function Evaluate(account, ABTestBeginning, workspaceA, workspaceB,
     var lossA = LossFunction(ordersA, sessionsA - ordersA, ordersB, sessionsB - ordersB),
         lossB = LossFunction(ordersB, sessionsB - ordersB, ordersA, sessionsA - ordersA)
 
-    var winner = ChooseWinner(ordersA, (bounceSessionsA + noBounceSessionsA) - ordersA, ordersB, (bounceSessionsB + noBounceSessionsB) - ordersB, boundError)    
+    var winner = ChooseWinner(ordersA, (bounceSessionsA + noBounceSessionsA) - ordersA, ordersB, (bounceSessionsB + noBounceSessionsB) - ordersB, boundError)
```
pedro823: Please use `let` or `const` instead of `var`.

TestCase.ts
```diff
@@ -1,12 +1,34 @@
...
-export async function ABtest(ctx: ColossusContext)
+const account = 'basedevmkp'
+
+export async function initializeABtest(ctx: ColossusContext)
 {
-    var account = 'basedevmkp'
-    var beginning = '2018'
-    const tResult = await Evaluate(account, beginning, 'master', 'abtesting', ctx)
+    const vbase = await new VBase(ctx.vtex)
+    var beginning = '2019-01-01' //formatDate(new Date().getTime())
```
pedro823: this could be `const`

TestCase.ts
```diff
@@ -1,12 +1,34 @@
...
+        name: testId,
+        timeStart: beginning
+    })
+}
+
+export async function ABTestStatus(ctx: ColossusContext)
+{
+    const vbase = await new VBase(ctx.vtex)
+
+    var data = await vbase.get('ABTest', 'currentTestAB.json')
```
pedro823: `'ABTest'` should be a global `const` in this file.

TestCase.ts
```diff
@@ -1,12 +1,34 @@
...
+    var beginning = '2019-01-01' //formatDate(new Date().getTime())
+    var testId = '0001'
 
+    await vbase.save('ABTest', 'currentTestAB.json', {
+        name: testId,
+        timeStart: beginning
+    })
+}
+
+export async function ABTestStatus(ctx: ColossusContext)
```
pedro823: since you're returning something, you should declare the return type

node/vbase.ts
```diff
@@ -0,0 +1,30 @@
+import { VBase as BaseClient} from '@vtex/api'
```
pedro823: _nit_: lint

node/vbase.ts
```diff
@@ -0,0 +1,30 @@
...
+  const readable = new Readable()
+  readable.push(JSON.stringify(arg))
+  readable.push(null)
+  return readable
+}
+
+export default class VBase {
+  client: BaseClient
+
+  constructor(opts: any) {
```
pedro823: `opts` should be typed to `BaseClient`'s constructor argument, not `any`

node/vbase.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.client = new BaseClient(opts)
+  }
+
+  get = async (bucket, path) => {
+    let file = await this.client.getFile(bucket, path)
+    return JSON.parse(file.data.toString())
+  }
+
+  save = async (bucket, path, data) => {
```
pedro823: _nit_:
```ts
public async save(bucket: string, path: string, data: any) {
	// ...
}
```

node/vbase.ts
```diff
@@ -0,0 +1,30 @@
...
+export default class VBase {
+  client: BaseClient
+
+  constructor(opts: any) {
+    this.client = new BaseClient(opts)
+  }
+
+  get = async (bucket, path) => {
+    let file = await this.client.getFile(bucket, path)
+    return JSON.parse(file.data.toString())
```
pedro823: `node-vtex-api` Already has [vbase.getJSON<T>](https://github.com/vtex/node-vtex-api/blob/master/src/VBase.ts#L58) and [vbase.saveJSON()](https://github.com/vtex/node-vtex-api/blob/master/src/VBase.ts#L70). You shouldn't need to code this again! You can find examples of use [over here](https://github.com/vtex/support-browser-session/blob/master/node/utils/oneTimeToken.ts#L20).

TestCase.ts
```diff
@@ -12,7 +11,8 @@ const testId = '0001'
 
 export async function initializeABtest(ctx: ColossusContext) {
     const vbase = await new VBase(ctx.vtex)
-    const beginning = formatDate(new Date().getTime())
+    const beginning = new Date().toISOString().substr(0, 10)
+    console.log(beginning)
```
pedro823: I don't think you need this `console.log`

ab-test/evaluate.ts
```diff
@@ -2,7 +2,7 @@ import axios from 'axios'
 import { ChooseWinner, LossFunction } from '../math-tools/decision-rule'
 import { ColossusContext } from 'colossus'
 
-const boundError = 0.00000001
+const boundError = 0.000000001
```
pedro823: You added a zero. Was it intentional? Why not write it in exponential form, like `1e-9`?

ab-test/evaluate.ts
```diff
@@ -26,8 +26,7 @@ export async function Evaluate(account, ABTestBeginning, workspaceA, workspaceB,
           lossB = LossFunction(orderSessionsB, noOrderSessionsB, orderSessionsA, noOrderSessionsA)
 
     let winner = ChooseWinner(orderSessionsA, noOrderSessionsA, orderSessionsB, noOrderSessionsB, boundError)
```
pedro823: This still could be const

ab-test/evaluate.ts
```diff
@@ -26,8 +26,7 @@ export async function Evaluate(account, ABTestBeginning, workspaceA, workspaceB,
...
-    if (winner == null)
-    {
+    if (winner == null) {
```
pedro823: does
```ts
if (!winner) {
```
work?

ab-test/evaluate.ts
tergol1: But as we could change the value (as in the part you commented `if (winner == null) { winner = ... } `) I think that it must not be a constant.

ab-test/evaluate.ts
tergol1: Yes! Thanks, a lot better

ab-test/evaluate.ts
tergol1: Yes, it was intentional! Didn't changed only because I'm lazy lol. It started in `0.001` and then added 0's

TestCase.ts
tergol1: You're right! I forgot this. This was just to assure the function `toISOString` did what I wanted

ab-test/evaluate.ts
pedro823: Oh, my bad.

pedro823: Although you could use
```ts
const winner = ChooseWinner(...) || 'not yet decided'
```

tergol1: Great suggestion again! Thanks5!

TestCase.ts
```diff
@@ -1,6 +1,5 @@
 import { Evaluate } from './ab-test/evaluate'
```
pedro823: _nit_: I think you should keep the folders in camel case.
```ts
import { Evaluate } from './abTest/evaluate'
```

TestCase.ts
```diff
@@ -21,9 +19,9 @@ export async function initializeABtest(ctx: ColossusContext) {
 }
 
 export async function ABTestStatus(ctx: ColossusContext): Promise<string> {
-    const vbase = await new VBase(ctx.vtex)
+    const { resources: { vbase } } = ctx
 
-    var data = await vbase.get(bucket, fileName)
+    let data = await vbase.get(bucket, fileName)
```
pedro823: what data is this? What is its type? Isn't there a better name to use?

By the way, `vbase.get` is a function that comes from `IODataSource` and not `VBase` and I wouldn't recommend using it directly. Since `data` is an object, why not call
`vbase.saveJSON` and `vbase.getJSON`?

Let's say you're storing an object called `ABTestData`, declared in `typings.d`. You'd use `vbase.saveJSON(bucket, fileName, abTestData)` and `vbase.getJSON<AbTestData>(bucket, fileName)`.

TestCase.ts
```diff
@@ -21,9 +19,9 @@ export async function initializeABtest(ctx: ColossusContext) {
...
     var beginning = data.timeStart
     if (data.name == testId) {
         beginning = data.timeStart
```
pedro823: This if is really not necessary anymore, is it?

ab-test/abtest-manager.ts
```diff
@@ -2,19 +2,19 @@ import { ColossusContext } from 'colossus'
...
+  ctx.body = 'A/B Test initialized'
+}
 
-export const  abTestStatus = async (ctx: ColossusContext) => {
-    ctx.set('Cache-Control', 'no-cache')
+export const abTestStatus = async (ctx: ColossusContext) => {
+  ctx.set('Cache-Control', 'no-cache')
 
-    var ResultAB = await ABTestStatus(ctx)
+  var ResultAB = await ABTestStatus(ctx)
```
pedro823: `const`

ab-test/abtest-manager.ts
```diff
@@ -2,19 +2,19 @@ import { ColossusContext } from 'colossus'
...
+  ctx.set('Cache-Control', 'no-cache')
 
-    await initialize(ctx)
+  await initialize(ctx)
 
-    ctx.status = 200
-    ctx.body = 'A/B Test initialized'
-  }
+  ctx.status = 200
+  ctx.body = 'A/B Test initialized'
```
pedro823: Question -- Do you need to leave the client hanging when you're awaiting for `initialize(ctx)`? Will you throw an error when it wasn't initialized? Because otherwise, you could launch this `await` as a promise that isn't waited. Example below:

```ts
export const initializeAbTest = async (ctx: ColossusContext) => {
  ctx.set('Cache-Control', 'no-cache')
  initialize(ctx).then(success => log.info(success))
                 .catch(err => log.error(err))

  ctx.status = 200
  ctx.body = 'A/B Test initialized'
```

node/index.ts
```diff
@@ -1,13 +1,58 @@
...
+import { initializeAbTest, abTestStatus } from '../ab-test/abtest-manager'
+import { Resources } from './resources/index'
+
+const prepare = (handler: any) => async (ctx: ColossusContext) => {
+  ctx.resources = new Resources(ctx)
+  try {
+    await handler(ctx)
+  } catch (err) {
+    if (err.code && err.message && err.status) {
+      console.log({ err })
```
pedro823: _future stuff_: You should log this error to splunk too.

[This PR kinda does it correctly, i guess](https://github.com/vtex/support-authority/pull/6/files). If you want to follow that direction. But that is _future stuff_.

node/resources/index.ts
```diff
@@ -0,0 +1,10 @@
+import { ColossusContext } from 'colossus'
+import VBase from '../vbase'
```
pedro823: Import it from `@vtex/api`. `yarn add @vtex/api`.

TestCase.ts
```diff
@@ -22,12 +22,12 @@ export async function ABTestStatus(ctx: ColossusContext): Promise<string> {
     const { resources: { vbase } } = ctx
 
     let data = await vbase.get(bucket, fileName)
-    var beginning = data.timeStart
-    if (data.name == testId) {
-        beginning = data.timeStart
+    if(data)
+    {
+        var beginning = data.timeStart
```
pedro823: `const`

TestCase.ts
```diff
@@ -22,12 +22,12 @@ export async function ABTestStatus(ctx: ColossusContext): Promise<string> {
...
+        const tResult = await Evaluate(account, beginning, 'master', 'abtesting', ctx)
+        console.log(tResult)
```
pedro823: remember to remove this if you don't need it, or log it to splunk

TestCase.ts
```diff
@@ -22,12 +22,12 @@ export async function ABTestStatus(ctx: ColossusContext): Promise<string> {
     const { resources: { vbase } } = ctx
 
     let data = await vbase.get(bucket, fileName)
-    var beginning = data.timeStart
-    if (data.name == testId) {
-        beginning = data.timeStart
+    if(data)
```
pedro823: Generally, we prefer to check edge cases instead of the happy paths, to avoid indentation and
[hadouken code](https://www.reddit.com/r/ProgrammerHumor/comments/27yykv/indent_hadouken/).
What about flipping this `if` and returning immediately if there is no data?

```ts
if (!data) {
  return 'Test not initialized'
}
```

ab-test/evaluate.ts
```diff
@@ -25,10 +25,7 @@ export async function Evaluate(account, ABTestBeginning, workspaceA, workspaceB,
     const lossA = LossFunction(orderSessionsA, noOrderSessionsA, orderSessionsB, noOrderSessionsB),
           lossB = LossFunction(orderSessionsB, noOrderSessionsB, orderSessionsA, noOrderSessionsA)
 
-    let winner = ChooseWinner(orderSessionsA, noOrderSessionsA, orderSessionsB, noOrderSessionsB, boundError)
-    if (winner == null) {
-        winner = 'not yet decided'
-    }
+    let winner = ChooseWinner(orderSessionsA, noOrderSessionsA, orderSessionsB, noOrderSessionsB, boundError) || 'not yet decided'
```
pedro823: well now you could've used `const`, haha.

TestCase.ts
```diff
@@ -12,7 +12,7 @@ export async function initializeABtest(ctx: ColossusContext) {
     const { resources: { vbase } } = ctx
     const beginning = new Date().toISOString().substr(0, 10)
 
-    await vbase.save('ABTest', 'currentTestAB.json', {
```
pedro823: _nit_: `currentABTest.json` is more english-like.

math-tools/statistics/bound-error.ts
```diff
@@ -0,0 +1,3 @@
+export const boundError = 1e-11
+
+// That won't be just a constant. We'll have a function to calculate it
```
pedro823: If it'll be a function, you can already prototype it:

```ts
export const boundError = (): number => 1e-11
```
Then everyone will already need to _call_ it to get a number.

ab-test/abtest-manager.ts
tergol1: No, I don't need it and that was again a nice suggestion! Thanks!

TestCase.ts
tergol1: It will be necessary later but due to a decision that I will make later I choose to delete it now.

tergol1:
Closing this PR. Continue discussion in next feature addition

PR #2, opened 2019-02-22
Adding Kullback-Leibler divergence Calculation

tergol1:
Add the Calculation of [Kullback-Leibler](https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence) divergence to help in identify if a test already has a winner.

We've identified that only the bound on error function is not fitted to claim that a test already has a winner. We'll use Kullback-Leibler and try to answer it more precisely.

#### Types of changes
* [ ]   Bug fix (non-breaking change which fixes an issue)
* [x]   New feature (non-breaking change which adds functionality)
* [ ]   Breaking change (fix or feature that would cause existing functionality to change)
* [ ]   Requires change to documentation, which has been updated accordingly.

### DIFF
--- a/mathTools/digamma-function.ts
+++ b/mathTools/digamma-function.ts
@@ -1,0 +1,145 @@
+// MODULES //
+export function evalpoly(P: number[], x)
+{
+    var num = 0
+
+    for(var i = 0; i < P.length; i++)
+    {
+        num += P[i] * Math.pow(x, i)
+    }
+
+    return num
+}
+
+export function evalrational(P: number[], Q: number[], x)
+{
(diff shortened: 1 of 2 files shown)


pedro823:
Did you copy this code from somewhere? What is its license?

If so, please create sanity-check tests to see if this is actually implementing the desired functions. The tests will be extremely useful, because then we can _beautify_ these functions knowing that we did not break them.

pedro823:
by the way, [there is this library I think you'll like.](http://mathjs.org/docs/reference/functions/kldivergence.html)

PR #3, opened 2019-03-08
Add a restriction on number of samples to the decision rule

tergol1:
To guarantee that observed data since beginning of test has statistical sufficiency.

The test previously could end without guaranteeing that the analysis was based on data that has statistical sufficiency.

Simulate a experiment where one variation is clearly better than other but with small number of samples, say 5000 (the reason for this value is found in code - it's small considering accuracy parameters predefined). The test should answer with `Not yet decided`.

#### Types of changes


abTest/requestMetrics/storedash.ts
```diff
@@ -1,10 +1,10 @@
 import axios from 'axios'
 
 const baseURL = 'http://api.vtex.com/api/storedash/'
-const metricsStoredashURL = '/metrics/storedash/SessionCube?from='
-const aggregationURL = '&to=now&operation=sum&aggregateBy=workspace,data.orders'
+const metricsStoredashURL = '/metrics/storedash/sessioncube?from='
```
iagoaraujo: This one can be only `metricsStoredashPath = '/metrics/storedash/sessioncube'`.
Also I think you can keep the query string in a function and receives the `from` parameter as an argument:
Ex:   aggregationQuery(from): string => ?from='${from}&to=now&operation=sum&fields=count&aggregateBy=workspace,data.orderPlaced'

tergol1: Is it that different from
```ts
const baseURL = 'http://api.vtex.com/api/storedash/'
const metricsStoredashURL = '/metrics/storedash/sessioncube?from='
const aggregationURL = '&to=now&operation=sum&fields=count&aggregateBy=workspace,data.orderPlaced'

export const StoreDashRequestURL = (account, ABTestBeginning): string => (
    baseURL + account + metricsStoredashURL + ABTestBeginning + aggregationURL)
```
besides the change of name?

tergol1: Changed my mind. I thought more and understood better the suggestion.

PR #5, opened 2019-03-27
Make AB-Tester log errors on Colossus

tergol1:
Every error should be logged on Colossus and this makes them searchable on Splunk.

Link AB-Tester and then look for a name of a workspace that doesn't. Then try to initialize the A/B test for this workspace. That will throw a error meaning that this workspace doesn't exists and saying what should be searched in Splunk to see the error. Then search it and see there the error.

#### Types of changes


node/abTest/commands/finish.ts
```diff
@@ -0,0 +1,20 @@
+import { FindWorkspace, TestingWorkspaces } from '../../workspace/list'
+import { FinishABTestMaster, FinishABTestParams } from '../../workspace/modify'
+import { firstOrDefault } from '../../utils/firstOrDefault'
+
+export async function FinishAbTestForWorkspace(ctx: ColossusContext): Promise<void> {
+    const { vtex: { account, route: { params: { workspace } } } } = ctx
```
iagoaraujo: *nit*: i'd use another name to this `workspace`, once it's not the current workspace where the requests is running and it can be a little confused.

node/index.ts
```diff
@@ -1,14 +1,20 @@
...
 import { timeToCompleteAbTest, initializeAbTestForWorkspace, abTestStatus, finishAbTestForWorkspace } from './abTest/manager';
 import { Resources } from './resources/index'
+import Logger from './utils/logger'
 
 const testManager = (handler: any) => async (ctx: ColossusContext) => {
   ctx.resources = new Resources(ctx)
   try {
     await handler(ctx)
   } catch (err) {
+    if (!err) {
```
iagoaraujo: Is it necessary? What is the scenario the err is undefined?

node/index.ts
```diff
@@ -1,14 +1,20 @@
...
+      err = {}
+    }
+    const logger = new Logger(ctx)
+
     if (err.code && err.message && err.status) {
-      console.log({ err })
+      logger.sendLog(err, err)
```
iagoaraujo: Why are you repeating the `err` as `details` too?
As details you can send what you think is important to debug the app.
Ex: a specific code that shows you this part of the code is with problem.

node/utils/logger.ts
```diff
@@ -0,0 +1,15 @@
+import { Colossus } from '@vtex/api'
```
iagoaraujo: I didn't find this module at node-vtex-api: https://github.com/vtex/node-vtex-api/tree/master/src
I think you need to import the `Logger`.

node/utils/logger.ts
```diff
@@ -0,0 +1,15 @@
...
+    private client: Colossus
+
+    constructor(private ctx: ColossusContext) {
+        this.client = new Colossus(ctx.vtex)
+    }
+
+    public sendLog = (level: LogLevel, data: object): Promise<void> => {
+        const { vtex: { account, userAgent, workspace } } = this.ctx
+        console.warn(`Try this query at Splunk to retrieve error log: 'index=colossus key=log_error subject=${userAgent} account=${account} workspace=${workspace}'`)
+        return this.client.sendLog(data, level)
```
iagoaraujo: I didn't find this method at node-vtex-api 🤔 
Is it right? It's the same problem with the Colossus import.

iagoaraujo:
Important: every request performed by the app should be treated as a "source" of problems, which means you always have to guarantee the error treatment with try/catch (and retry if you think it's necessary). I noticed some vbase and router access are just requests that you assume you'll have no problem with them. You should always catch the errors and log them.

node/index.ts
tergol1: Actually I think that it happened once when I was testing the changes and then I put this but as I cannot even remember and less imagine a scenario like that, I'll remove it

tergol1:
> Important: every request performed by the app should be treated as a "source" of problems, which means you always have to guarantee the error treatment with try/catch (and retry if you think it's necessary). I noticed some vbase and router access are just requests that you assume you'll have no problem with them. You should always catch the errors and log them.

I have a doubt about what you just said. First, none observation about lack of error treatment in requests to `vbase` and I'm correcting it now. But about those to router I thought that, as the methods are extensions of those defined on `Workspaces` class in `vtex@api` they inherited its error treatment. Also, I'm in doubt about where else I should treat errors in requests and I'm not figuring out.

node/index.ts
iagoaraujo: Up

node/abTest/commands/finish.ts
```diff
@@ -0,0 +1,23 @@
+import { FindWorkspace, TestingWorkspaces } from '../../workspace/list'
```
iagoaraujo: *nit*: I think it's better rename the file `workspace/list` to `workspace/workspaces`
Also I noticed the method FindWorkspace first asks for the list and then iterate in the list to search the workspace. Why don't perform a GET to `/<account>/<workspace>` and check if the workspace exists? This endpoint probably will return 404.

node/abTest/commands/finish.ts
```diff
@@ -0,0 +1,23 @@
...
+import { firstOrDefault } from '../../utils/firstOrDefault'
+
+export async function FinishAbTestForWorkspace(ctx: ColossusContext): Promise<void> {
+    const { vtex: { account, route: { params: { finishingWorkspace } } } } = ctx
+
+    try {
+        const workspaceName = firstOrDefault(finishingWorkspace)
+        if (!FindWorkspace(account, workspaceName, ctx)) {
+            ctx.body = `workspace "${finishingWorkspace}" doesn't exists.`,
+                ctx.status = 401
```
iagoaraujo: 401? If the workspace doesn't exists I think you should return a Bad Request, or maybe Not Found (this one is better)

node/abTest/commands/initialize.ts
```diff
@@ -0,0 +1,24 @@
...
+import { firstOrDefault } from '../../utils/firstOrDefault'
+import { initializeABtest } from '../saveBeginning'
+
+export async function InitializeAbTestForWorkspace(ctx: ColossusContext): Promise<void> {
+    const { vtex: { account, route: { params: { probability, initializingWorkspace } } } } = ctx
+    const workspaceName = firstOrDefault(initializingWorkspace)
+    try {
+        if (!FindWorkspace(account, workspaceName, ctx)) {
+            ctx.body = `workspace "${workspaceName}" doesn't exists.`,
+                ctx.status = 401
```
iagoaraujo: Same thing here

node/abTest/commands/finish.ts
```diff
@@ -0,0 +1,23 @@
...
+        }
+
+        var testingWorkspaces = await TestingWorkspaces(account, ctx)
+        if (testingWorkspaces.length == 2) {
+            await FinishABTestMaster(account, ctx)
+        }
+        await FinishABTestParams(account, workspaceName, ctx)
+    } catch (err) {
+        throw new Error(err)
```
iagoaraujo: You can set the response status code when you catch the error.
`try` --> everything is ok, nothing wrong. In case of something unexpected, you can throw the error and let the catch set the status code of error.
`catch` --> in case of errors of something unexpected, log the error e set the status code if you know the reason of the error.
So if the error is Not Found, you can assume that it probably didn't find the workspace and set the response as you want.

(also remember to log the error)

node/abTest/commands/initialize.ts
```diff
@@ -0,0 +1,24 @@
...
+        }
+
+        var testingWorkspaces = await TestingWorkspaces(account, ctx)
+        if (!testingWorkspaces) {
+            await InitializeABTestMaster(account, ctx)
+        }
+        await InitializeABTestParams(account, workspaceName, ctx)
+        await initializeABtest(Number(probability), ctx)
+    } catch (err) {
```
iagoaraujo: check the comment at `finish.ts`

node/abTest/commands/status.ts
```diff
@@ -0,0 +1,22 @@
...
+        const data = await vbase.get(bucket, fileName)
+        if (!data) {
+            return [DefaultEvaluationResponse('Test not initialized', 'none', 'none')]
+        }
+        const beginning = data.timeStart
+        const probability = data.probability
+
+        return await TestWorkspaces(account, beginning, probability, ctx)
+    } catch (err) {
+        throw new Error(err)
```
iagoaraujo: Log the error

node/abTest/commands/timeToComplete.ts
```diff
@@ -0,0 +1,10 @@
+import { TimeToCompleteAbTest } from '../time-to-complete'
+
+export async function TTCAbTest(ctx: ColossusContext): Promise<number> {
+    const { vtex: { account, route: { params: { probability } } } } = ctx
+    try {
+        return await TimeToCompleteAbTest(account, Number(probability), ctx)
+    } catch (err) {
+        throw new Error(err)
```
iagoaraujo: Log

node/abTest/commands/finish.ts
tergol1: The only problem with rename the file is that already exists one called `workspaces`, which extends the class `Workspace` from `vtex/api` to include the `ABTestParameters`. I agree with the name, I just thought to await which nowadays is `workspaces` to be implemented in workspace promoter app.

The GET is clearly a simple and good optimization, changing it now.

node/abTest/commands/finish.ts
```diff
@@ -0,0 +1,23 @@
...
+import { firstOrDefault } from '../../utils/firstOrDefault'
+import { LoggerClient as Logger } from '../../clients/logger'
+
+export async function FinishAbTestForWorkspace(ctx: ColossusContext): Promise<void> {
+    const { vtex: { account, route: { params: { finishingWorkspace } } } } = ctx
+    const workspaceName = firstOrDefault(finishingWorkspace)
+    try {
+        await FinishABTestParams(account, workspaceName, ctx)
+        var testingWorkspaces = await TestingWorkspaces(account, ctx)
+        console.log(testingWorkspaces)
```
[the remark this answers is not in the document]
tergol1: Already removed but I forgot to commit. Sorry

victorges:
After adding the file to `git`, you need to explicit remove it to stop git from tracking the file, even if it is now matching the `.gitignore`. 

So try running this `git rm node_modules` (or with some flag for recursiveness idk)

victorges:
Sorry, `CMD+Enter` apparently closes the PR 😶

victorges:
Also please squash this before merging or `git` will probably still keep the `node_modules` in the history

tergol1:
> Sorry, `CMD+Enter` apparently closes the PR 😶

I was laughing at the notification without understanding why the PR had been closed.

PR #6, opened 2019-04-17
Do not try to update ABTestParameters when there is none workspaces being tested

tergol1:
Do not allow `ab-tester` to execute unnecessary processes.

Before this modification, every time the app listen the periodic event `pulse` from colossus, it will execute all the process to update `ABTestParameters`, which is unnecessary.

#### Types of changes


node/abTest/commands/update.ts
```diff
@@ -10,16 +10,17 @@ export async function UpdateStatusOnEvent(ctx: EventsContext): Promise<void> {
   const { account, resources: { vbase } } = ctx
 
   try {
-    const data = await vbase.get(bucket, fileName)
-    let beginning = data.dateOfBeginning
-    if (beginning === undefined) {
-      beginning = new Date()
-    }
-
     const testingWorkspaces = await TestingWorkspaces(account, ctx)
```
victorges: Does this function have any caching inside? Not a change in this PR but I believe we will need to add some caching later to avoid doing a request to the DB on every account pulse

tergol1: I agree. Starting think about this change immediately and try to code as early as possible.

PR #7, opened 2019-04-24
Add verification to recent data gathered before use it to update ABTestParameters

tergol1:
Detect awkward situations for two compared workspaces where theirs conversion rate don't behave like it's expected and then act restarting the A/B Test Parameters aiming to redistribute the traffic in the workspaces to check if the recent data observed can predict the actual conversion rate of some workspaces. That last part is important as we don't want to Restart the parameters if the verification returns `false` for any workspace (which means that for this workspace the data is ok) and we should keep the test flow normally. If at least one workspace has the test result correct for it, at least this workspace will have the answer if it's better or worse than `master`.

Events like seasonality of visitors can distort metrics in situations where a test is being executed and the traffic of workspaces has already been modified. E.g. a distribution of 60:40 traffic can give the workspace with more traffic a benefit when it comes a period where better visitors arrive as they're distributed 60:40 between them.
This PR should allow the traffic control part of the code detect when something like that is happening and redirect traffic to verify if the change in a metric should be treated as an anomaly due to specific events that don't corresponds to the general behavior of a visitor.

It's kind of difficult to actually test this thing in production-like environment (it depends on such a awkward situation that the verification for a Restart of workspace's traffic parameters returns `true`) but it can be tested for sanity in isolated environment.

Only took the code relative to the function `RandomRestart` and then:
 - Generate fake data of workspaces with types `WorkspaceData` and `WorkspaceCompleteData` such that conversion rate differs a lot from `SinceBeginning` to `Last24Hours` - also the relative variation which can be achieved with relative variations with opposite signal;
 - Call `RandomRestart` for these `WorkspaceCompleteData` generated some times and see if the function returns true.

#### Types of changes

P.S.: This PR contains tons of refactoring and code cleaning including changes in names of files and variables. The main focus here is the file `node/workspace/modify.ts` and the function `RandomRestart`.


tergol1:
[ch9843]

node/abTest/test-workspaces.ts
```diff
@@ -1,39 +1,38 @@
...
 export async function TestWorkspaces(account: string, abTestBeginning: string, probability: number, ctx: ColossusContext): Promise<TestResult[]> {
     const testingWorkspaces = await TestingWorkspaces(account, ctx.vtex)
-    const beginningQuery = HoursSince(abTestBeginning)
-    const workspacesData = await FilterWorkspacesData(account, beginningQuery, testingWorkspaces, ctx.vtex)
     const Results: TestResult[] = []
-    if (workspacesData.length > 0) {
-        await UpdateWorkspacesData(account, beginningQuery, testingWorkspaces, ctx.vtex)
-        const workspacesCompleteData = await BuildCompleteData(account, ctx, workspacesData)
+    if (testingWorkspaces.length > 0) {
+        const beginningQuery = MinutesSinceQuery(abTestBeginning)
```
igorframos: Can this be in a separate function so it is easier to read?

node/abTest/test-workspaces.ts
```diff
@@ -1,39 +1,38 @@
...
+        const workspacesData = await FilterWorkspacesData(account, beginningQuery, testingWorkspaces, ctx.vtex)
+        if (workspacesData.length > 0) {
+            const workspacesCompleteData = await BuildCompleteData(account, abTestBeginning, ctx.vtex, workspacesData)
```
igorframos: This could also be a function, I guess.

node/abTest/test-workspaces.ts
```diff
@@ -1,39 +1,38 @@
...
+                    Results.push(await Evaluate(abTestBeginning, masterWorkspace, workspaceData, probability))
+                }
             }
         }
-    }
-    else {
-        for (const workspaceName of testingWorkspaces) {
-            if (workspaceName !== 'master') {
-                Results.push(DefaultEvaluationResponse(abTestBeginning, 'master', workspaceName))
+        else {
```
igorframos: The else is very short, so you could probably invert your condition, have the short code with an early return and reduce the indentation of the rest.

node/clients/vbase.ts
```diff
@@ -1,6 +1,12 @@
 import { VBase as BaseClient } from '@vtex/api'
 import { Readable } from 'stream'
 
+const bucketName = 'ABTest'
+const fileName = 'currentABTest.json'
+
+// TODO: `testId` should be determined in a general way based on account and workspaces
+const testId = '0001'
```
igorframos: For now, just concatenate everything and, maybe, hash it. It will be better than having a fixed ID.

node/clients/vbase.ts
```diff
@@ -31,4 +37,17 @@ export default class VBase {
...
+    const beginning = new Date().toISOString().substr(0, 16)
+    return await this.save(bucketName, fileName, {
+      Id: testId,
+      dateOfBeginning: beginning,
+      probability: (probability),
+    } as ABTestData)
+  }
+
+  public finishABtest = async (): Promise<void> => {
+    await this.client.deleteFile(bucketName, fileName)
```
igorframos: Do we want a history of the A/B tests that have been performed? Maybe we can just move this info somewhere? I think it is too much to do it now, but I would welcome a `TODO` here with this idea.

node/clients/vbase.ts
tergol1: The next PR solve this problem because we use a "unique" identifier and the one related to who knows this info and how to use it.

node/clients/vbase.ts
tergol1: You touched a bigger problem that is the history about A/B tests. I think that some informations about a test should be saved but we must think carefully about it because maybe this involve some use of this data that can affect our clients, e.g. use clustered info about retailers of some kind of product.

tergol1: But maybe we just want to save data about each client and use just for him and I don't know if it represents any problem. Maybe discuss during next weekly

PR #8, opened 2019-05-17
Use ABTesting API

tergol1:
Make `ab-tester` use new ABTesting API to get, save and delete data about workspaces being tested.

We create this new API for two reasons:
 - We don't want anymore to save metadata for A/B test in our database (probably we want to explode it) and now we think it should be better to save it in Metadata (now, after we will move it to Chronos).
 - Use this new API that save and read from Metadata and then already has cache implemented in its responses.

The second one should grant us a performance gain.

Call the routes defined by the app to initialize a test for a workspace and then call the ABTesting API get routes to verify if the data was saved successfully. Repeat this for updating this data and deleting.

[ch11234]

#### Types of changes


node/typings.d.ts
```diff
@@ -53,4 +53,17 @@ declare global {
     abTestParameters: ABTestParameters
     production: boolean
   }
+
+  export interface WorkspacesMetadata {
```
victorges: **nit**: Indentation seems a bit off here

node/typings.d.ts
```diff
@@ -53,4 +53,17 @@ declare global {
...
+    Id: string,
+    workspaces: TestingWorkspace[]
+}
+export interface TestingWorkspace {
```
victorges: **nit**: Add a blank line between declarations

node/utils/parametersToUpdate.ts
```diff
@@ -0,0 +1,13 @@
+export const Parameters = (workspacesData: Map<string, WorkspaceData>, workspaces: string[]): ParametersMetadata[] => {
+    const parameters: ParametersMetadata[] = []
+    for (const workspace of workspaces) {
+        if (workspacesData.has(workspace)) {
+            parameters.push({
+                a: workspacesData.get(workspace)!.OrderSessions,
```
victorges: What is this `!.`? (typescript noob)

node/utils/workspace.ts
```diff
@@ -108,4 +108,67 @@ const ToABTestParameters = (workspaceData: WorkspaceData): ABTestParameters => {
         a: workspaceData.OrderSessions + 1,
         b: workspaceData.NoOrderSessions + 1,
     }
+}
+
+export const WorkspaceNameToTesting = (workspace: string): TestingWorkspace => {
```
victorges: I think all these helpers could be implemented beautifully with some functional library like `ramda`. Perhaps ask for suggestions/review from someone that uses typescript daily like @tlgimenes or @vcalasans  👀

victorges: Also, I think this function might actually be a constructor for the `TestingWorkspace` type.

node/utils/workspaceData.ts
```diff
@@ -0,0 +1,7 @@
+export const MapWorkspaceData = (workspacesData: WorkspaceData[]): Map<string, WorkspaceData> => {
```
victorges: ramdaaaa

node/utils/workspace.ts
```diff
@@ -165,8 +165,7 @@ export const InitialParameters = (workspaces: TestingWorkspace[]): ParametersMet
     const parameters: ParametersMetadata[] = []
     for (const workspace of workspaces) {
         parameters.push({
-            a: 1,
-            b: 1,
+            abTestParameters: InitialABTestParameters,
```
victorges: Is this `InitialABTestParameters` a function? Are you missing an actual function call?

node/utils/workspace.ts
```diff
@@ -103,7 +103,7 @@ export const ToWorkspaceMetadada = (workspaceData: WorkspaceData, weight: number
     }
 }
 
-const ToABTestParameters = (workspaceData: WorkspaceData): ABTestParameters => {
+export const ToABTestParameters = (workspaceData: WorkspaceData): ABTestParameters => {
```
victorges: This seems perhaps a constructor as well

node/abTest/commands/status.ts
```diff
@@ -6,9 +6,12 @@ const bucket = 'ABTest'
 const fileName = 'currentABTest.json'
 
 export async function ABTestStatus(ctx: ColossusContext): Promise<TestResult[]> {
-  const { vtex: { account }, resources: { vbase } } = ctx
+  const { vtex: { account }, resources: { router, vbase } } = ctx
 
   try {
+    if(await router.getWorkspaces(account) === null) {
```
victorges: **nit**: spacing

node/abTest/commands/status.ts
```diff
@@ -6,9 +6,12 @@ const bucket = 'ABTest'
...
+      throw new Error('Test not initialized')
```
victorges: Can you return a specific error status code here, like a `400` or a `404`?

node/typings/testingParameters.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+const UnmapParameters = (mapParameters: Map<string, ParametersMetadata>): ParametersMetadata[] => {
+    const parameters: ParametersMetadata[] = []
+    for (const parameter of mapParameters.values()) {
+        parameters.push(parameter)
+    }
+    return parameters
+}
+
+declare global {
```
victorges: What is this `global`? If it is what it sounds (like declaring something anyone could use without importing this file), can we avoid it?

tergol1: I used it to create some arrays of this type but when coding I felt like a bad decision but didn't came out with anything better. I'll try to put some effort to make this part better.

node/typings/testingParameters.ts
```diff
@@ -0,0 +1,54 @@
...
+        return UnmapParameters(this.parameters)
+    }
+
+    public Includes = (workspaceName: string): boolean => {
+        return this.parameters.has(workspaceName)
+    }
+
+    public Set = (workspacesData: Map<string, WorkspaceData>) => {
+        for (const workspace of this.parameters.keys()) {
+            if (workspacesData.has(workspace)) {
```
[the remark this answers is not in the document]
tergol1: The only detail here that bothered me here is that the already saved workspaces are the ones that we know that are being tested. The data that comes from `Storedash` contains other workspaces, e.g. `undefined`, and we don't want to add these. So I iterated over these workspaces that I know for sure that are the only being tested and updated theirs data. Also I think that inverting the iteration order, i.e. iterate over `workspacesData ` and verifies if the workspace is being tested, adds a verification operation.

node/utils/workspace.ts
tergol1: It's a constant declared in the first line. Do you think I should change it's name? Or other suggestion aiming clarify the code? I thought it was ok because it was a proper name so kind of reminds me of a constant's name.

node/workspace/modify.ts
```diff
@@ -23,7 +23,9 @@ export async function FinishABTestParams(account: string, workspace: string, ctx
 async function UpdateABTestParams(account: string, workspaceData: WorkspaceData, ctx: IOContext) {
     const workspaces = await new ABWorkspaces(ctx)
     const abWorkspace = await workspaces.get(account, workspaceData.Workspace)
-    await workspaces.set(account, abWorkspace.name, ToWorkspaceMetadada(workspaceData, abWorkspace.weight, abWorkspace.production))
+    if(Math.min(workspaceData.NoOrderSessions, workspaceData.OrderSessions) >= 1) {
```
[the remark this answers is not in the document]
tergol1: Sorry about that! haha
Can I neglect this suggestion as this piece of code will die very soon?

node/utils/guid.ts
```diff
@@ -0,0 +1,9 @@
+export default class Guid {
+    public static newGuid() {
```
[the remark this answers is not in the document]
tergol1: Great! I will!

node/abTest/analysis/time/timeToComplete.ts
```diff
@@ -1,17 +1,16 @@
-import { IOContext } from '@vtex/api'
-import { GetWorkspacesData, StoreDashRequestURL } from '../../../clients/storedash'
+import { GetWorkspacesData } from '../../../clients/storedash'
```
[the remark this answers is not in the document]
tergol1: I thought about doing something that maybe would be better than what I did and solve this problem: implement this as a method inside the class.

node/clients/storedash.ts
```diff
@@ -25,8 +25,8 @@ async function getDataFromStoreDash(endPoint: string, ctx: IOContext): Promise<J
     })
 }
 
-export async function GetWorkspacesData(endPoint: string, ctx: IOContext): Promise<WorkspaceData[]> {
-    const metrics = await getDataFromStoreDash(endPoint, ctx)
+export async function GetWorkspacesData(from: string, storedash: Storedash): Promise<WorkspaceData[]> {
```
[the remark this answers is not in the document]
tergol1: There is your suggestion that is the same thing that I commented in a previous comment haha
Sorry! Yeah, that should be and will do it now.

node/abTest/commands/initialize.ts
```diff
@@ -1,18 +1,32 @@
...
+            await router.setWorkspaces(account, testingWorkspaces)
         }
+        testingWorkspaces.workspaces = AddWorkspaceToTest({ name: workspaceName }, testingWorkspaces.workspaces)
+        await router.setWorkspaces(account, testingWorkspaces)
+
+        await InitializeParameters(account, testingWorkspaces.workspaces, router)
+
         await InitializeABTestParams(account, workspaceName, ctx.vtex)
-        await vbase.initializeABtest(1 - Number(probability))
+        await vbase.initializeABtest(1 - Number(probability), Id)
```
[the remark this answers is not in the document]
tergol1: That was part of a previous plan about storing the previous results but I changed my mind during the process and didn't make the other things needed. The plan was to store two different datas: that one that we use during the test and another one that we only write anything when the test is finished. Well, the last part was never done. I think that it could be removed.

node/utils/parametersToUpdate.ts
tergol1: That I guarantee that this isn't `undefined`.

node/clients/logger.ts
```diff
@@ -1,4 +1,4 @@
-import { InstanceOptions, IOContext, Logger } from '@vtex/api'
+import { InstanceOptions, IOContext, Logger, LogLevel } from '@vtex/api'
```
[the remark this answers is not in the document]
tergol1: I used to assign the correct type of a parameter of the function `sendLog`.

node/abTest/commands/initialize.ts
```diff
@@ -1,18 +1,32 @@
...
+            testingWorkspaces = {
+                Id: (Id),
+                workspaces: AddWorkspaceToTest({ name: 'master' }, []),
+            }
+            await router.setWorkspaces(account, testingWorkspaces)
         }
+        testingWorkspaces.workspaces = AddWorkspaceToTest({ name: workspaceName }, testingWorkspaces.workspaces)
+        await router.setWorkspaces(account, testingWorkspaces)
+
+        await InitializeParameters(account, testingWorkspaces.workspaces, router)
```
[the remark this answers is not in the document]
tergol1: The `InitializeABTestParams` is to initialize the parameters by the old way modifying the parameter's values stored in the db.

PR #9, opened 2019-06-10
Improve Confidence Level Calculation

tergol1:
Implement the calculation of significance level.

We want a better Level calculation which also must be dynamic.

To understand the calculation, read https://en.wikipedia.org/wiki/P-value .

Call route for status and check the presence of the field `PValue`.

#### Types of changes


node/abTest/commands/initialize.ts
```diff
@@ -16,6 +15,7 @@ export async function InitializeAbTestForWorkspace(ctx: ColossusContext): Promis
         if (!hasTestingWorkspaces ? true : !testingWorkspaces.Includes('master')) {
             await InitializeABTestParams(account, 'master', ctx.vtex)
             testingWorkspaces.Add('master')
+            const uuidv4 = require('uuid/v4')
```
victorges: You should always use typescript imports rather than `require`, as requires cannot be statically parsed by the compiler/webpack/whatevs

node/abTest/data/buildData.ts
```diff
@@ -10,7 +10,7 @@ export async function BuildCompleteData(account: string, abTestBeginning: string
     const workspaceCompleteData = new Map()
     for (const workspaceNewData of workspacesNewData) {
         if(testingWorkspaces.Includes(workspaceNewData.Workspace)) {
-            workspaceCompleteData[workspaceNewData.Workspace] = CompleteData(GetWorkspaceData(workspacesData, workspaceNewData.Workspace), workspaceNewData)
+            workspaceCompleteData.set(workspaceNewData.Workspace, CompleteData(GetWorkspaceData(workspacesData, workspaceNewData.Workspace), workspaceNewData))
```
victorges: Why this change?

node/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "vtex.ab-tester",
-  "version": "0.8.1",
+  "version": "0.9.0",
   "dependencies": {
-    "axios": "^0.18.0",
-    "ramda": "^0.25.0"
+    "ramda": "^0.25.0",
+    "uuid": "3.3.2"
```
victorges: Do you need to fix the version for the lib? Keeping a `^` would probably be better

node/utils/mathTools/statistics/samplesRestrictions.ts
```diff
@@ -13,4 +15,9 @@ export function NumberOfSamples(boundError: number, BoundProbability: number): n
 
 export const SamplesRestriction = (WorkspaceA: WorkspaceData, WorkspaceB: WorkspaceData, boundError: number, boundProbability: number): boolean => {
     return WorkspaceA.Sessions + WorkspaceB.Sessions > NumberOfSamples(boundError, boundProbability)
+}
+
+export const pValue = (control: ABTestParameters, alternative: ABTestParameters): number => {
+    const controlConversion = (control.a - 1) / (control.a + control.b - 2)
+    return Math.pow(controlConversion, alternative.a - 1) * Math.pow(1 - controlConversion, alternative.b - 1) * beta(alternative.a, alternative.b)
```
victorges: image

tergol1: HAHAHA that piece of code should change a little over the next few days but the point here is just implement the function and add this info to status.

node/abTest/data/buildData.ts
tergol1: Something exploded in the previous way when publishing the app.
More precisely, the error was something like "Type Map<any, any> has no iterator"

node/package.json
tergol1: The missing `^` was just my fault. The fix here was just me being dumb and importing it in the wrong place.

PR #11, opened 2019-06-13
Remove confidence level

tergol1:
Remove the use of the confidence level inputed by the user.

We don't need anymore to request  the desired confidence level and then we can remove all its uses.

Call `initialize` and `status` and verify if `ab-tester` works properly, i.e. initialize test correctly and return response for `status` correctly.

#### Types of changes


node/service.json
```diff
@@ -7,7 +7,7 @@
       "public": false
     },
     "initializeAbTestForWorkspace": {
-      "path": "/_v/private/abtesting/initialize/:initializingWorkspace/:probability",
+      "path": "/_v/private/abtesting/initialize/:initializingWorkspace",
```
victorges: Won't this break all existing clients? If there are any systems in production calling the old API, it'd be better to do a change now only _adding_ the new API while still proxying the old one to the same code and ignoring the last parameter.

node/utils/mathTools/forBetaDistribution/decisionRule.ts
```diff
@@ -31,20 +33,23 @@ export function LossFunctionChossingVariantOne(Beta1: ABTestParameters, Beta2: A
...
+    const betaA = WorkspaceToBetaDistribution(WorkspaceA)
+    const betaB = WorkspaceToBetaDistribution(WorkspaceB)
+    const pvalue = pValue(betaA, betaB)
 
-    if (!SamplesRestriction(WorkspaceA, WorkspaceB, BoundError, delta)) {
-        return null
-    }
-    if (chooseA && chooseB) {
+    const chooseA = LossFunctionChossingVariantOne(betaA, betaB) < epsilon
+    const chooseB = LossFunctionChossingVariantOne(betaB, betaA) < epsilon
```
victorges: Much better with the intermediate variables :)

node/utils/mathTools/forBetaDistribution/decisionRule.ts
```diff
@@ -31,16 +34,22 @@ export function LossFunctionChossingVariantOne(Beta1: ABTestParameters, Beta2: A
...
+    const betaB = WorkspaceToBetaDistribution(WorkspaceB)
+    const pvalue = pValue(betaA, betaB)
+
+    const chooseA = LossFunctionChossingVariantOne(betaA, betaB) < epsilon
+    const chooseB = LossFunctionChossingVariantOne(betaB, betaA) < epsilon
+    const draw = pvalue > Confidence
+    const distinct = pvalue < 1 - Confidence
 
-    if (chooseA && chooseB) {
+    if (chooseA && chooseB && draw) {
```
victorges: I'm not sure about this, but perhaps did you mean something like `((chooseA == chooseB) || draw)`?

victorges: Clarified offline

PR #15, opened 2019-07-03
Create logs for ab-tester methods

tergol1:
Add logs to every time someone initialize or finishes a A/B Test and also when is requested the status of the test.

Check in Splunk if logs are being sent.

#### Types of changes
- [x] Auditability


node/clients/logger.ts
```diff
@@ -2,20 +2,28 @@ import { InstanceOptions, IOContext, Logger, LogLevel } from '@vtex/api'
...
-export class LoggerClient {
+export default class LoggerClient {
   private client: Logger
 
   constructor(private ctx: IOContext, opts: InstanceOptions) {
     this.client = new Logger(ctx, opts)
   }
 
-  public sendLog = (level: LogLevel, message: string | {}): Promise<void> => {
+  public sendMethodLog = (message: string | {}, method: string): Promise<void> => {
```
victorges: This could be named simply `info`, as what is present in the `node-vtex-api`'s `Logger` client as well. What do you think of having both a `message` field which is a string and a `data` field that is a key/value (I guess this `{}` type), and you interpolate the data with the message, adding the fixed `message` field with the string?

This is similar to what we do in our `go` logging library, and the good side is that it encourages always having *typed* logs, as well as a human-friendly message with what that log means.

node/abTest/commands/initialize.ts
```diff
@@ -31,12 +30,12 @@ export async function InitializeAbTestForWorkspace(ctx: ColossusContext): Promis
         })
 
         await vbase.initializeABtest(ctx.vtex)
+        logger.sendMethodLog(`A/B Test initialized in ${account} for workspace ${workspaceName}`, 'TestInitialized')
```
victorges: It is better to have typed logs whenever possible, like having specific fields with this account and workspace so we can search for that later on splunk without the need for some kind of regex

node/clients/logger.ts
```diff
@@ -2,20 +2,28 @@ import { InstanceOptions, IOContext, Logger, LogLevel } from '@vtex/api'
...
+    const data = (typeof message === 'string' || message instanceof String)
+      ? { message }
+      : message
+
+    return this.client.sendLog(method, data, LogLevel.Info)
+  }
+
+  public sendErrorLog = (message: string | {}): Promise<void> => {
     const { account, userAgent } = this.ctx
     console.warn(`Try this query at Splunk to retrieve error log: 'index=colossus key=log_error subject=${userAgent} account=${account} workspace=${this.ctx.workspace}'`)
```
victorges: This log/warn seems unnecessary

node/clients/logger.ts
```diff
@@ -2,20 +2,28 @@ import { InstanceOptions, IOContext, Logger, LogLevel } from '@vtex/api'
...
-  public sendLog = (level: LogLevel, message: string | {}): Promise<void> => {
+  public sendMethodLog = (message: string | {}, method: string): Promise<void> => {
+    const data = (typeof message === 'string' || message instanceof String)
+      ? { message }
+      : message
+
+    return this.client.sendLog(method, data, LogLevel.Info)
+  }
+
+  public sendErrorLog = (message: string | {}): Promise<void> => {
```
victorges: Similarly, this could be simply `error`

node/index.ts
```diff
@@ -1,18 +1,15 @@
...
 const tester = (handler: any) => async (ctx: ColossusContext) => {
   ctx.resources = new Resources(ctx.vtex)
   try {
     await handler(ctx)
   } catch (err) {
-    const logger = new Logger(ctx.vtex, {})
-
     if (err.code && err.message && err.status) {
-      logger.sendLog(err, { status: ctx.status, message: err.message })
+      ctx.logger.sendErrorLog(err, { status: ctx.status, message: err.message })
```
victorges: I think you are sending 2 arguments to a function that only receives 1 🤔

tergol1: Yup, you're right. Thanks for spotting this.

PR #16, opened 2019-07-15
Custom parameters of initialization

tergol1:
Allow custom configuration of duration of initial stage and the proportion of traffic in each workspace during this part of the test.

We regularly faced the problem of customize the way the initial stage of the test runs, i.e. how long it takes and the proportion of traffic directed to each workspaces. This PR creates a new route that configures custom parameters that aims to allow these configurations.

It's funny but this is being tested in production with Boticario. Just start a test with custom parameters and see if we have the desired behavior.

#### Types of changes


node/abTest/analysis/time/initialStage.ts
```diff
@@ -1,14 +1,29 @@
...
+}
+
+export const InitialParameters = (proportion: number, workspaces: ABTestWorkspace[]): Map<string, WorkspaceData> => {
+    const map = new Map<string, WorkspaceData>()
+    let noOrderSessions = 0
+    let orderSessions = 0
+
+    for (const workspace of workspaces) {
+        [noOrderSessions, orderSessions] = workspace.name !== MasterWorkspaceName ? [0, 0]
+            : [(1 - proportion) * INFIINITE_MASS, proportion * INFIINITE_MASS]
```
igorframos: s/INFIINITE_MASS/INFINITE_MASS/g

PR #19, opened 2019-08-21
ab tests meta data memory opt

lucas-daher:
Keep the ab tester in pace with the changes in the Kube router to optimise memory usage when saving the ab tests metadata.

Memory usage optimisation.

Building an app and testing if it works with the kube router after the changes (that version is ready for pulling)

#### Types of changes

### DIFF
--- a/node/abTest/commands/finish.ts
+++ b/node/abTest/commands/finish.ts
@@ -18,17 +18,13 @@ export async function FinishAbTestForWorkspace(ctx: ColossusContext): Promise<vo
       return
     }
 
-    const testingParameters = new TestingParameters(testingWorkspaces.ToArray())
+    const testingParameters = new TestingParameters(workspaceMetadata.workspaces)
+    testingParameters.Remove(workspaceName)
     await router.setWorkspaces(account, {
       id: workspaceMetadata.id,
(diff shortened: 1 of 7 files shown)


node/abTest/commands/finish.ts
```diff
@@ -23,10 +23,15 @@ export async function FinishAbTestForWorkspace(ctx: ColossusContext): Promise<vo
       id: workspaceMetadata.id,
       workspaces: testingWorkspaces.ToArray(),
     })
-    const tsmap = new TSMap<string, Workspace>([...testingParameters.Get()])
+    
+    let tsmap = new TSMap<string, ABTestParameters>()
+    for (let entry of testingParameters.Get()) {
```
tergol1: Commenting here but it demands changes in other places:
I think would be better to change the class `TestingParameters` to don't store anymore a map of `<string, Workspace>`; we should use a map of `<string, ABTestParameters>`. This should makes the class lighter and then its creation will  also be optimized. Doing this will avoid this repetition of this code with an iteration to create this map which will be used in PUT method to save the parameters the router's API.

lucas-daher: solved

PR #20, opened 2019-08-23
adding orders value to ab metrics

lucas-daher:
Add "orders value" (revenue) to ab tests metrics

Adding another metric for clients to use when analysing ab tests (adding it to the tool belt in the near future) and later to be used as the main metric if the client wishes to do so. 

Running the ab tester linked to some ab test and checking the stats.

#### Types of changes

### DIFF
--- a/node/utils/evaluation-response.ts
+++ b/node/utils/evaluation-response.ts
@@ -15,6 +15,8 @@ export const DefaultEvaluationResponse = (abTestBeginning: string, workspaceANam
     WorkspaceB: workspaceBName,
     WorkspaceBSessions: 0,
     WorkspaceBSessionsLast24Hours: 0,
+    WorkspaceAOrdersValue: 0,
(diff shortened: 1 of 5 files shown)


node/utils/evaluation-response.ts
```diff
@@ -34,4 +36,6 @@ export const EvaluationResponse = (abTestBeginning: string, workspaceAData: Work
     WorkspaceB: workspaceBData.SinceBeginning.Workspace,
     WorkspaceBSessions: workspaceBData.SinceBeginning.Sessions,
     WorkspaceBSessionsLast24Hours: workspaceBData.Last24Hours.Sessions,
+    WorkspaceAOrdersValue: workspaceAData.SinceBeginning.OrdersValue,
```
tergol1: What about also add the revenue in the last 24 hours?

PR #21, opened 2019-09-18
Feature/weighted dist

lucas-daher:
Change AB test parameters to be passed to the router as weights of a uniform distribution. The conversion from beta to uniform is done with the ProbabilityYBeatsAll function.

Changes the meaning of the parameters will make the router faster, because the sampling process is going to be simpler.

Linking this version and testing against new of prod versions of the router (for roll out and possible roll back)

#### Types of changes


tergol1:
What about change the minor as this is not a bug fix:

node/utils/mathTools/forBetaDistribution/decisionRule.ts
```diff
@@ -21,6 +21,71 @@ export function ProbabilityOfOneBeatTwo(a: number, b: number, c: number, d: numb
...
+                CurrX.push(X[j])
+                idxs.push(0)
+            }
+        }
+        curr += ProbabilityAllBeatY(Y, CurrX, idxs, 0)
+        curr *= Math.pow(-1, idxs.length)
+        ret += curr
+    }
+    return {a:ret, b:1}
+}
```
tergol1: **nit** I'm trying to think in a way to made this calculation more like a step-by-step solution of a math problem. This suggestion is for the code between lines 24 and 86.

- I'm ok with `ProbabilityAllBeatY`, really well written code!
- I'm not getting the point of `NextComb` (maybe I'm the problem lol). I could only think about something like the following code to generate an array of combinations with size `p`

```
export function Combinations(set: any[], p: number): any[] {
    let combinations = []
    let current = []
    let tailcombs = []

    if (p > set.length || p <= 0) {
        return []
    }

    if (p === set.length) {
        return [set]
    }

    if (p === 1) {
        combinations = []
        for (const elem of set) {
            combinations.push(elem)
        }
        return combinations
    }

    for (let i = 0; i < set.length - p + 1; i++) {
        current = set.slice(i, i + 1)
        tailcombs = Combinations(set.slice(i + 1), p - 1)
        for (const elem of tailcombs) {
            combinations.push(current.concat(elem))
        }
    }
    return combinations
}
```
 With that code, we could execute for each combination (inside a `for` loop varying the quantity `p`) `ProbabilityAllBeatY `.

Explaining what I didn't understand:

```
Recursively NextComb applied to [0, 0, 0]
[0, 0, 0] -> [1, 0, 0] -> [2, 1, 0] -> [3, 1, 0] -> [4, 1, 0] -> [5, 1, 0] -> [6, 1, 0] -> ...

Comment (lucas-daher) [reply node/utils/mathTools/forBetaDistribution/decisionRule.ts]: About NextComb, it actually goes like this:

[0, 0, 0] -> [1, 0, 0] -> [0, 1, 0] -> [1, 1, 0] -> [0, 0, 1] -> [1, 0, 1] -> [1, 1, 1]

and is hopefully (and guaranteed by an invariant) to never be called after this.

I change the i++ to the next line to make it clearer.

File: node/typings/testingParameters.ts [hunk 2]
```diff
@@ -29,11 +30,19 @@ export default class TestingParameters {
     }
 
     public Set = (workspacesData: Map<string, WorkspaceData>) => {
+        let names = Array<string>(0)
+        let beta_params = Array<ABTestParameters>(0)
```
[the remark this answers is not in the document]
lucas-daher: I'm trying to remember to always use camelCase, but whenever I'm not paying attention, I'll just default back to snake_case.

node/abTest/updateParameters.ts
```diff
@@ -19,12 +19,12 @@ export async function UpdateParameters(account: string, aBTestBeginning: string,
...
-        
         const tsmap = new TSMap<string, ABTestParameters>([...testingParameters.Get()])
-        await router.setParameters(account, {
+        router.setParameters(account, {
             Id: testId,
             parameterPerWorkspace: tsmap,
         })
+        
         return
     }
```
lucas-daher: @tergol1 I couldn't figure out how to comment in lines under this one, but in line 40 maybe we should remove the `await`

node/utils/mathTools/forBetaDistribution/decisionRule.ts
tergol1: I just saw the order of `++` and now I see that I thought about the algorithm as `++i`, but you actually implemented `i++`. Now I'm ok with both solutions, separating or not the index increment.

node/abTest/updateParameters.ts
tergol1: Yes, we should! Great remark!

PR #22, opened 2019-10-01
Fix/initializing parameters

tergol1:
Fix the way fixed parameters are initialized.

After change the way `ABTestParameters` are set - which came from a new purpose for its values - we broke the API to initialize a test with fixed parameters as the one for proportion of traffic didn't make sense anymore.

Try to initialize with parameters, get status and finish a A/B Test.

#### Types of changes


node/abTest/commands/initialize.ts
```diff
@@ -33,17 +33,18 @@ async function InitializeAbTest(hoursOfInitialStage: number, proportionOfTraffic
...
+
         const tsmap = new TSMap<string, ABTestParameters>([...testingParameters.Get()])
         await router.setParameters(account, {
             Id: workspaceMetadata.id,
             parameterPerWorkspace: tsmap,
         })
 
         await vbase.initializeABtest(hoursOfInitialStage, proportionOfTraffic, ctx.vtex)
-        logger.info(`A/B Test initialized in ${account} for workspace ${workspaceName}`, { account: `${account}`, workspace: `${workspaceName}`, method: 'TestInitialized' })
+        logger.info(`A/B Test initialized in ${account} for workspace ${workspaceName}`, { account: `${account}`, workspace: `${workspaceName}`, proporiton: `${proportionOfTraffic}`, method: 'TestInitialized' })
```
pedro823: typo: `proporiton` should be `proportion`

node/typings/testingParameters.ts
```diff
@@ -30,20 +32,34 @@ export default class TestingParameters {
...
         for (const workspace of this.parameters.keys()) {
             if (workspacesData.has(workspace)) {
                 names.push(workspace)
                 betaParams.push(WorkspaceToBetaDistribution(workspacesData.get(workspace)!))
             }
         }
-        for (let i = 0; i < betaParams.length; i++) {
-            let y = betaParams.shift()!
+        for (const _ of betaParams) {
+            y = betaParams.shift()!
```
pedro823: I don't think you're gaining performance for declaring the `y` variable in the start of the function instead of declaring a `const` here. The V8 engine for js is _stupidly_ good at optimizing this, and you've got a layer of `tsc` (probably with `babel` as well) that may as well optimize it.

node/typings/testingParameters.ts
```diff
@@ -30,20 +32,34 @@ export default class TestingParameters {
...
+
         for (const workspace of this.parameters.keys()) {
             if (workspacesData.has(workspace)) {
                 names.push(workspace)
                 betaParams.push(WorkspaceToBetaDistribution(workspacesData.get(workspace)!))
             }
         }
-        for (let i = 0; i < betaParams.length; i++) {
-            let y = betaParams.shift()!
+        for (const _ of betaParams) {
```
pedro823: I really don't like that you're iterating through something that is being changed inside the loop... Is it a guarantee that js's iterator follows the list changes?

node/typings/testingParameters.ts
```diff
@@ -30,20 +32,34 @@ export default class TestingParameters {
...
+            y = betaParams.shift()!
             this.parameters.set(names.shift()!, ProbabilityYBeatsAll(y, betaParams))
             betaParams.push(y)
         }
     }
+
+    public SetWithFixedParameters = (proportion: number) => {
+        const actualProportion = proportion > 1 && proportion < 10000 ? proportion : 10000
```
pedro823: if proportion = 0, `actualProportion` will be set to `10000`. Is that the intended behaviour?

node/typings/testingParameters.ts
```diff
@@ -30,20 +32,34 @@ export default class TestingParameters {
...
+        const size = this.parameters.size
+        const nonMasterParameter = (10000 - actualProportion) / (size - 1)
+        let parameter = 0
+
+        for (const workspace of this.parameters.keys()) {
+            parameter = workspace === MasterWorkspaceName ? proportion : nonMasterParameter
```
pedro823: Same thing as before, I don't think you're gaining performance with declaring `parameter` before this loop.

node/typings/testingParameters.ts
```diff
@@ -30,20 +32,34 @@ export default class TestingParameters {
...
+        for (const _ of betaParams) {
+            y = betaParams.shift()!
             this.parameters.set(names.shift()!, ProbabilityYBeatsAll(y, betaParams))
             betaParams.push(y)
         }
     }
+
+    public SetWithFixedParameters = (proportion: number) => {
+        const actualProportion = proportion > 1 && proportion < 10000 ? proportion : 10000
+        const size = this.parameters.size
```
pedro823: judging by the constructor, this looks like it could be `undefined`.

node/typings/testingParameters.ts
tergol1: Nice!

node/typings/testingParameters.ts
tergol1: Yes! We have an interval of values that should contain the value of `proportion`. If this is not the case, we should send traffic only to `master`.

lucas-daher: Trying to understand this clearly: `proportion` is the proportion of traffic that's going to master, correct?

Maybe we should accept numbers including 0 and 10000. Any reasons not to?

node/typings/testingParameters.ts
lucas-daher: I don't think so, `parameters` is never `null`. If everything goes wrong, it's constructed as an empty map.

node/typings/testingParameters.ts
lucas-daher: Do you really want to use `proportion` here? To be consistent with the previous lines, I think this should be:

`parameter = workspace === MasterWorkspaceName ? actualProportion : nonMasterParameter`

node/typings/testingParameters.ts
tergol1: I agree with you. I'll change.

node/typings/testingParameters.ts
tergol1: Yes! Thanks for spotting it!

node/typings/testingParameters.ts
```diff
@@ -34,29 +34,29 @@ export default class TestingParameters {
...
+        for (let i = 0; i < size; i++) {
+            const y = betaParams.shift()!
             this.parameters.set(names.shift()!, ProbabilityYBeatsAll(y, betaParams))
             betaParams.push(y)
         }
     }
 
     public SetWithFixedParameters = (proportion: number) => {
-        const actualProportion = proportion > 1 && proportion < 10000 ? proportion : 10000
+        const actualProportion = proportion > 0 && proportion < 10001 ? proportion : 10000
```
pedro823: `proportion` could be `10000.999`, best to use `proportion <= 10000`. Leads to better legibility as well.

node/typings/testingParameters.ts
pedro823: parameters is never null, however `parameters.size` could be undefined, no?

node/typings/testingParameters.ts
tergol1: Actually it can't be and even if it was, that's not a problem as will be converted later to 10000. And why do you think it leads to better legibility?

pedro823: Because js's `Number` is always floating point, I always prefer to not leave any ambiguity and write `>= x` instead of `> x - 1`. The only exception is when i'm working with indices of an array. But that is _completely_ preference

node/typings/testingParameters.ts
lucas-daher: `.size` is defined in `Map` interface, and always returns `number`, so... no.

node/typings/testingParameters.ts
lucas-daher: nit: would change `proportion > 0` to `proportion >= 0`. Current behaviour might confuse the user .

PR #23, opened 2019-11-18
Feat/save abtests results

lucas-daher:
Save the results of all AB tests done.

Currently the data is erased when the test is finished and we have no way of keeping track of tests performed, results etc.

Link the ab tester;
Start an ab test;
Finish the test;
Check if the data was saved.

#### Types of changes

A couple of things:
- I'm not sure if what I did was good practice (see comment in the code);
- I have no idea how to create automated tests for this. For the ab tester overall, actually. But will look into it soon.


node/clients/vbase.ts
```diff
@@ -27,25 +28,43 @@ export default class VBase {
...
+    }
+    catch {
+      //does nothing, if this causes an error, probably just means resultsList.json desn't exist yet and will be created
+    }
+    finally {
+      list.push(testResults)
+      await this.save(results, testResults, ctx)
+      await this.save(list, resultsList, ctx)
+      console.log("check point111")
+    }
```
lucas-daher: Any better ideas on how to do this?

node/clients/vbase.ts
```diff
@@ -27,25 +28,40 @@ export default class VBase {
...
+      list = JSON.parse(listFile.data.toString()) as string[]
+    }
+    catch {
+      //does nothing, if this causes an error, probably just means resultsList.json desn't exist yet and will be created
+    }
+    finally {
+      list.push(testResults)
+      await this.save(results, testResults, ctx)
+      await this.save(list, resultsList, ctx)
+    }
```
lucas-daher: Any better ideas on how to do this?

pedro823: I think the best idea is to separate the fetcher of the file in a different function, and then use something like
```js
const testResults = [...await this.fetchTestResults(ctx.account), testResultsFile]
await Promise.all([this.save(results, testResultsFile, ctx), this.save(testResults, resultsList, ctx)])
```
(Assume that you renamed `testResults` to `testResultsFile`)

That way, you:
- Have smaller, neater code
- Do both `save` calls asynchronously "at the same time"

node/abTest/commands/finish.ts
```diff
@@ -14,9 +14,12 @@ export async function FinishAbTestForWorkspace(ctx: ColossusContext): Promise<vo
...
       await router.deleteParameters(account)
       await router.deleteWorkspaces(account)
-      let beginning = (await vbase.get(ctx.vtex)).dateOfBeginning
-      if (beginning === undefined) {
+      let beginning = ""
+      const data = await vbase.get(ctx.vtex)
+      if (!data || data.dateOfBeginning === undefined) {
         beginning = new Date().toISOString().substr(0, 16)
+      } else {
+        beginning = data.dateOfBeginning
```
tergol1: **nit** WDYT about ternary operator?

```
const beginning = (!data || data.dateOfBeginning === undefined) ?
                  new Date().toISOString().substr(0, 16) :
                  data.dateOfBeginning
```

node/clients/vbase.ts
```diff
@@ -27,25 +28,40 @@ export default class VBase {
...
       initialStageTime: initialTime,
-    } as VBaseABTestData, ctx)
+    } as VBaseABTestData, fileName, ctx)
   }
 
-  public finishABtest = async (ctx: IOContext): Promise<void> => {
+  public finishABtest = async (ctx: IOContext, results: TestResult[]): Promise<void> => {
+    
     await this.client.deleteFile(bucketName(ctx.account), fileName)
+    let testResults = 'TestResults' + results[0].ABTestBeginning + '.json'
```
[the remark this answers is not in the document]
tergol1: Good question! I think this should be checked by this point even if following the logic here this should always exists.

node/clients/vbase.ts
```diff
@@ -27,25 +27,27 @@ export default class VBase {
...
       initialProportion: proportion,
       initialStageTime: initialTime,
-    } as VBaseABTestData, ctx)
+    } as VBaseABTestData, fileName, ctx)
   }
 
-  public finishABtest = async (ctx: IOContext): Promise<void> => {
+  public finishABtest = async (ctx: IOContext, results: TestResult[]): Promise<void> => {
     await this.client.deleteFile(bucketName(ctx.account), fileName)
+    let file = 'TestResults' + results[0].ABTestBeginning + '.json'
```
tergol1: **nit** `const file = 'TestResults' + results[0].ABTestBeginning + '.json'`

tergol1: **nit** You can make this a const in this file like `bucketName` as 
```
const testResultsFileName = (results: TestResult[]) => `TestResults${results[0].ABTestBeginning}.json`
```

node/clients/vbase.ts
```diff
@@ -27,25 +27,27 @@ export default class VBase {
     }
   }
 
-  public save = async (data: any, ctx: IOContext) => {
+  public save = async (data: any, file: string, ctx: IOContext) => {
```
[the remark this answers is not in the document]
lucas-daher: I like the name as is for a few reasons:
- this is on a different abstraction level than the `saveFile` functions, so there should be no confusion;
- this name is consistent with the `get` function;

node/clients/vbase.ts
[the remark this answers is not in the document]
lucas-daher: great point, it actually isn't.

node/clients/vbase.ts
```diff
@@ -27,25 +28,40 @@ export default class VBase {
...
 
-  public finishABtest = async (ctx: IOContext): Promise<void> => {
+  public finishABtest = async (ctx: IOContext, results: TestResult[]): Promise<void> => {
+    
     await this.client.deleteFile(bucketName(ctx.account), fileName)
+    let testResults = 'TestResults' + results[0].ABTestBeginning + '.json'
+    let list: string[] = []
+    try {
+      const listFile = await this.client.getFile(bucketName(ctx.account), resultsList)
+      list = JSON.parse(listFile.data.toString()) as string[]
```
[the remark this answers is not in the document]
lucas-daher: TBH, I don't really know how this works, just copied a similar line from the `get` function

node/clients/vbase.ts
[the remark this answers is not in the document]
tergol1: I agree with both points but I would like to think about both names also being changed. But this is just my opinion and you're free to ignore the suggestion.

PR #24, opened 2019-11-25
Feature/testing params interface

lucas-daher:
Create a testing parameters generic interface, so we can create different testing parameters classes to be used in different AB tests. Currently, only one class is created, for the test using conversion as main metric. Also starts saving test type in vbase and adds an optional parameter to `InitializeAbTest`.

This makes way for adding revenue as test metric by allowing the creation of different classes implementing `ITestingParameters`.

2 tests:

Test 1:
- Link this version of the app
- Start an AB test
- Check if everything works (all the tester routes and Colossus periodic update route)
- Unlink the app
- Link a new one from master
- Check if everything still works

Test 2:
- Start an AB test
- Check if everything works (all the tester routes and Colossus periodic update route)
- Link this version of the app
- Check if everything still works
 
#### Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)


node/typings/testingParameters.ts
```diff
@@ -69,12 +81,12 @@ const MapInitialParameters = (workspaces: ABTestWorkspace[]): Map<string, ABTest
     }
     return map
 }
-const UnmapParameters = (mapParameters: Map<string, ABTestParameters>): ABTestParameters[] => {
-    const parameters: ABTestParameters[] = []
-    for (const parameter of mapParameters.values()) {
-        parameters.push(parameter)
+
+export function createTestingParameters(testType: TestType): ITestingParameters {
+    if (testType == TestType.conversion) {
```
ruybrito106: **nit**: this seems like a good opportunity for a ternary :awesome:

node/typings/testingParameters.ts
```diff
@@ -1,12 +1,32 @@
...
+    Update(workspacesData: Map<string, WorkspaceData>): void;
+
+    UpdateWithFixedParameters(proportion: number): void;
+}
+
+export class TestingParametersConversion implements ITestingParameters{
     private parameters: Map<string, ABTestParameters>
 
-    constructor(testingWorkspaces: ABTestWorkspace[]) {
+    constructor() {
```
ruybrito106: it looks like the constructor and the `Init` methods have the same purpose, right? doesnt it make sense to join them? if there are multiple use cases with different initilization params, maybe you could make them optional.

node/typings/testingParameters.ts
```diff
@@ -1,12 +1,32 @@
...
+
+    Get(): Map<string, ABTestParameters>;
+
+    Add(workspaceName: string): void;
+
+    Remove(workspaceName: string): void;
+
+    Update(workspacesData: Map<string, WorkspaceData>): void;
+
+    UpdateWithFixedParameters(proportion: number): void;
```
pedro823: *nit*: `camelCase`

node/typings/testingParameters.ts
pedro823: even though the program's returning the same thing inside the `if` and outside?

node/typings/testingParameters.ts
```diff
@@ -69,12 +81,12 @@ const MapInitialParameters = (workspaces: ABTestWorkspace[]): Map<string, ABTest
     }
     return map
 }
-const UnmapParameters = (mapParameters: Map<string, ABTestParameters>): ABTestParameters[] => {
-    const parameters: ABTestParameters[] = []
-    for (const parameter of mapParameters.values()) {
-        parameters.push(parameter)
+
+export function createTestingParameters(testType: TestType): ITestingParameters {
```
pedro823: *nit*: `export const createTestingParameters = (testType: TestType): ITestingParameters => new TestingParametersConversion()`

pedro823:
oops

node/typings/testingParameters.ts
ruybrito106: 😕

lucas-daher: I know this part of the code doesn't make much sense now, but other `if`s will be added in the near future (at least one more).

PR #25, opened 2019-11-27
Fix/vbase client

tergol1:
This PR aims to adapt the last change we made in `ab-tester` evolving how it stores the data about previous tests in a way that allow us to use it also as an index to the the files with test's metadata. Also it adds some logs for possible errors when trying to read data from VBase.

The path we're following is one that allow us to don't delete anymore the metadata about the tests and for that we want to create some index that can be searched to find what is the current test, if is there one.

The main change relates to the way we store the test results. Now we'll save the index as an interface - which is a format we could evolve now and maybe in the future if we want - and then stores there the array with previous tests and also a field that says what is the current test being executed.

We should try to initialize and finish an A/B Test and in every state-transition we should check the files where we save the metadata to verify if they are changed accordingly to what we expect.

#### Types of changes


node/clients/vbase.ts
```diff
@@ -2,8 +2,8 @@ import { IOContext, VBase as BaseClient } from '@vtex/api'
 import { Readable } from 'stream'
 
 const bucketName = (account: string) => 'ABTest-' + account
-const fileName = 'currentABTest.json'
-const resultsListFile = 'resultsList.json'
+const abTestHistoryFile = 'abTestHistory.json'
+const testFileName = 'currentABTest.json'
```
pedro823: Does this break compatibility with the other version?

node/clients/vbase.ts
```diff
@@ -19,54 +19,103 @@ export default class VBase {
...
+      const rawDataHistory = await this.get(abTestHistoryFile, ctx)
+      testHistory = JSON.parse(rawDataHistory.data.toString()) as ABTestHistory
+    } catch (ex) {
+      ctx.logger.error(ex)
+      testHistory = {
+        finishedTests: [],
+        onGoing: '',
+      } as ABTestHistory
     }
+    return testHistory
```
pedro823: You don't need the variable `testHistory` at all. Just return the value :)

node/abTest/commands/status.ts
```diff
@@ -9,7 +9,7 @@ export async function ABTestStatus(ctx: ColossusContext): Promise<TestResult[]>
       ctx.status = 400
       throw new Error('Test not initialized')
     }
-    const data = await vbase.get(ctx.vtex)
+    const data = await vbase.getTestData(ctx.vtex)
```
pedro823: What about renaming the variable to `testData`?

node/abTest/commands/update.ts
```diff
@@ -8,7 +8,7 @@ export async function UpdateStatusOnEvent(ctx: EventsContext): Promise<void> {
     const workspacesMetadata = await router.getWorkspaces(account)
     const testingWorkspaces = new TestingWorkspaces(workspacesMetadata)
     if (testingWorkspaces.Length() > 0) {
-      const data = await vbase.get(ctx)
+      const data = await vbase.getTestData(ctx)
```
pedro823: ditto

node/clients/vbase.ts
```diff
@@ -19,54 +19,103 @@ export default class VBase {
...
     const beginning = new Date().toISOString().substr(0, 16)
-    return this.save({
-      dateOfBeginning: beginning,
-      initialProportion: proportion,
-      initialStageTime: initialTime,
-    } as VBaseABTestData, fileName, ctx)
+    const testHistory = await this.fetchTestHistory(ctx)
+
+    try {
+      this.save({
```
pedro823: this needs to be awaited, doesn't it?

node/clients/vbase.ts
```diff
@@ -19,54 +19,103 @@ export default class VBase {
...
+        initialStageTime: initialTime,
+      } as VBaseABTestData, testFileName, ctx)
+    } catch (ex) {
+      ctx.logger.error(ex)
+      throw new Error(`Something went wrong while initiliazing the test!`)
+    }
+
+    testHistory.onGoing = beginning
+    try {
+      this.save(testHistory, abTestHistoryFile, ctx)
```
pedro823: same here

By the way, can't you do the saves at the same time?

```ts
try {
   await Promise.all([this.save(...), this.save(...)])
}
```

node/clients/vbase.ts
```diff
@@ -19,54 +19,103 @@ export default class VBase {
...
+    } catch (ex) {
+      ctx.logger.error(ex)
+      await this.client.deleteFile(bucketName(ctx.account), testFileName)
```
pedro823: What if this also fails? Are we gonna stay in a inconsistent state?

node/clients/vbase.ts
```diff
@@ -19,54 +19,103 @@ export default class VBase {
...
   public finishABtest = async (ctx: IOContext, results: TestResult[]): Promise<void> => {
-    await this.client.deleteFile(bucketName(ctx.account), fileName)
+    const testHistory = await this.fetchTestHistory(ctx)
+    testHistory.onGoing = ''
+    testHistory.finishedTests.push(results[0].ABTestBeginning)
+    if (testHistory.finishedTests.length > 100) {
+      testHistory.finishedTests.shift()
+    }
+    try {
+      this.save(testHistory, abTestHistoryFile, ctx)
```
pedro823: this also needs to be awaited. Unless you're not caring about the exception, you need to await.

node/clients/vbase.ts
```diff
@@ -19,54 +19,103 @@ export default class VBase {
...
+      await this.save(results, testResultsFile, ctx)
     }
   }
 
-  private fetchResultsList = async (ctx: IOContext): Promise<string[]> => {
+  private get = async (file: string, ctx: IOContext) => {
     try {
-      const listFile = await this.client.getFile(bucketName(ctx.account), resultsListFile)
-      return JSON.parse(listFile.data.toString()) as string[]
+      return this.client.getFile(bucketName(ctx.account), file)
```
pedro823: This also needs to be awaited. Even though `return await` looks horrible.

node/clients/vbase.ts
tergol1: Just stop saving into the old one and start to save into the new.

node/clients/vbase.ts
tergol1: It's true haha
Thanks!

node/clients/vbase.ts
tergol1: Yes! Good catch!

node/clients/vbase.ts
lucas-daher: I don't think so, but this will make us lose some test results forever, like tear drops in the rain... or until someone remembers to look at the `resultsList.json` file. And this was up for so little time that probably no ab tests were done.

node/clients/vbase.ts
```diff
@@ -19,48 +19,91 @@ export default class VBase {
...
+  public getTestData = async (ctx: IOContext): Promise<VBaseABTestData> => {
     try {
-      const file = await this.client.getFile(bucketName(ctx.account), fileName)
-      return JSON.parse(file.data.toString()) as VBaseABTestData
+      const testData = await this.get(testFileName, ctx)
+      return testData as VBaseABTestData
     } catch (ex) {
-      throw new Error(`Get request for key ${fileName} in bucket ${bucketName(ctx.account)} failed!`)
+      ctx.logger.error(ex)
+      throw new Error(`Get request for key ${testFileName} in bucket ${bucketName(ctx.account)} failed!`)
```
lucas-daher: is this necessary here? `this.get` throws the same error

node/clients/vbase.ts
```diff
@@ -19,48 +19,91 @@ export default class VBase {
...
+        initialStageTime: initialTime,
+      } as VBaseABTestData, testFileName, ctx)
+    } catch (ex) {
+      ctx.logger.error(ex)
+      throw new Error(`An error occurred initializing the test!`)
+    }
+
+    testHistory.onGoing = beginning
+    try {
+      await this.save(testHistory, abTestHistoryFile, ctx)
```
lucas-daher: why not save both at the same time?

node/clients/vbase.ts
```diff
@@ -19,48 +19,91 @@ export default class VBase {
...
+  private fetchTestHistory = async (ctx: IOContext): Promise<ABTestHistory> => {
+    try {
+      const abTestHistory = await this.get(abTestHistoryFile, ctx) as ABTestHistory
+      return abTestHistory
+    } catch (ex) {
+      ctx.logger.error(ex)
+      return {
+        finishedTests: [],
+        onGoing: '',
+      } as ABTestHistory
```
lucas-daher: is this a good idea? If something goes wrong and we return this, the original file might me overwritten and all the information lost (this time forever like teardrops in the rain)

tergol1: I was thinking about that also... I accepted this option in face of the tradeoff between this and  getting an error while trying to execute any test operation. But now I changed my mind and do prefer to return an error.

node/clients/vbase.ts
tergol1: Doing in a known order we control where a possible error occurs and then is easier to maintain consistence in stored data - if something goes wrong in the last save, we can delete the previous one as we know for sure that nothing wrong happened with its save operation.

node/clients/vbase.ts
tergol1: You're right!

node/clients/vbase.ts
tergol1: Great spot! We shouldn't accept an test that starts with inconsistent metadata. I just committed changes for this. Thanks!

lucas-daher:
A couple of comments:

Food for thought: I don't really see any good reason for keeping `onGoing` at `ABTestHistory`: the data is read only twice and not really used. This creates a non trivial down side - possible inconsistency issues - and no real upside (at least for now). I would favor keeping data about the ongoing test only at `currentABTest.json`. This also makes it the indisputable source of truth about current tests.

As you mentioned, we will have a problem when deploying the code. idk how to work around it.
One terrible possibility is to create an empty file by hand in every account.
How likely is for the fetch to fail if the file exists? Are there different error messages depending on what happens? If there are different errors, we could return an empty object only in case the file doesn't exist. If the other possibilities of failing are really, really unlikely, I'd also feel more comfortable doing it.

node/abTest/commands/finish.ts
```diff
@@ -14,11 +14,11 @@ export async function FinishAbTestForWorkspace(ctx: Context): Promise<void> {
     if (testingWorkspaces.Length() <= 1) {
       await abTestRouter.deleteParameters(account)
       await abTestRouter.deleteWorkspaces(account)
-      const data = await storage.get(ctx.vtex)
+      const data = await storage.getTestData(ctx)
```
pedro823: Whoops, there's this that could be renamed to `testData` as well.

node/abTest/commands/status.ts
```diff
@@ -9,11 +9,11 @@ export async function ABTestStatus(ctx: Context): Promise<TestResult[]> {
       ctx.status = 400
       throw new Error('Test not initialized')
     }
-    const data = await storage.get(ctx.vtex)
-    if (!data) {
+    const testData = await storage.getTestData(ctx)
```
pedro823: this throws if it doesn't work, right?

node/clients/vbase.ts
pedro823: This is still relevant, right?

node/clients/vbase.ts
```diff
@@ -13,48 +13,108 @@ const jsonStream = (arg: any) => {
...
+    try {
+      await this.save(testHistory, abTestHistoryFile, ctx)
+    } catch (ex) {
+      ctx.clients.logger.error(ex)
+      try {
+        await this.deleteFile(bucketName(ctx.vtex.account), testFileName)
+      } catch (ex) {
+        ctx.clients.logger.error({exception: ex, error: 'inconsistent_state'})
+        throw new Error(`An error occurred initializing the test and its metadata are inconsistent!`)
+      }
```
pedro823: I feel this could be a helper function so that there's less indentation & `try/catch` in the same place.

node/clients/vbase.ts
```diff
@@ -13,48 +13,108 @@ const jsonStream = (arg: any) => {
...
+      await Promise.all([this.save(testHistory, abTestHistoryFile, ctx), this.save(results, testResultsFile, ctx)])
+    } catch (ex) {
+      ctx.clients.logger.error(ex)
+      throw new Error(`Something went wrong while finishing the test and updating its metadata!`)
+    }
+
+    try {
+      await this.deleteFile(bucketName(ctx.vtex.account), testFileName)
+    } catch (ex) {
+      ctx.clients.logger.error({exception: ex, error: 'inconsistent_state'})
```
pedro823: About these logs of inconsistent state -- you should probably log the _most_ things you can, so that manual fixing will be easier. We hope that those will be a small percentage, so don't worry about the log size. Log the account and workspace, or maybe even the `ctx`.

PR #26, opened 2019-11-29
Update client

tergol1:
Update client imported from `node-vtex-api`.

Test **every** `ab-tester` method.

#### Types of changes
- [x] Update classes imported from libs


node/abTest/commands/initialize.ts
```diff
@@ -1,26 +1,25 @@
...
 import { v4 as uuid } from 'uuid'
-import Router from '../../clients/router'
 import TestingParameters from '../../typings/testingParameters'
 import TestingWorkspaces from '../../typings/testingWorkspace'
 import { firstOrDefault } from '../../utils/firstOrDefault'
 
-export async function InitializeAbTestForWorkspace(ctx: ColossusContext): Promise<void> {
+export async function InitializeAbTestForWorkspace(ctx: Context): Promise<void> {
     await InitializeAbTest(1, 0.5, ctx)
 }
```
pedro823: Can't this function just return the void promise instead of awaiting?

```ts
export function InitializeAbTestForWorkspace(ctx: Context): Promise<void> {
    return InitializeAbTest(1, 0.5, ctx)
}
```

node/abTest/commands/initialize.ts
```diff
@@ -1,26 +1,25 @@
...
 
-export async function InitializeAbTestForWorkspaceWithParameters(ctx: ColossusContext): Promise<void> {
+export async function InitializeAbTestForWorkspaceWithParameters(ctx: Context): Promise<void> {
```
pedro823: ditto for this function, it could just return the void promise.

node/abTest/commands/initialize.ts
```diff
@@ -54,8 +53,8 @@ async function InitializeAbTest(hoursOfInitialStage: number, proportionOfTraffic
     }
 }
 
-async function InitializeWorkspaces(account: string, id: string, testingWorkspaces: ABTestWorkspace[], router: Router): Promise<void> {
-    await router.setWorkspaces(account, {
+async function InitializeWorkspaces(ctx: Context, account: string, id: string, testingWorkspaces: ABTestWorkspace[]): Promise<void> {
+    await ctx.clients.abTestRouter.setWorkspaces(account, {
```
pedro823: do you need to pass the context down if you're just getting the `abTestRouter`?

node/abTest/commands/initialize.ts
```diff
@@ -54,8 +53,8 @@ async function InitializeAbTest(hoursOfInitialStage: number, proportionOfTraffic
     }
 }
 
-async function InitializeWorkspaces(account: string, id: string, testingWorkspaces: ABTestWorkspace[], router: Router): Promise<void> {
-    await router.setWorkspaces(account, {
+async function InitializeWorkspaces(ctx: Context, account: string, id: string, testingWorkspaces: ABTestWorkspace[]): Promise<void> {
```
pedro823: this function could also just return the promise.

node/abTest/updateParameters.ts
```diff
@@ -13,22 +11,23 @@ import { InitializeParameters } from './initializeParameters'
 
 const MasterWorkspaceName = 'master'
 
-export async function UpdateParameters(account: string, aBTestBeginning: string, hoursOfInitialStage: number, proportionOfTraffic: number,
-    workspacesData: WorkspaceData[], testingWorkspaces: TestingWorkspaces, testId: string, router: Router, storedash: Storedash): Promise<void> {
+export async function UpdateParameters(ctx: Context, account: string, aBTestBeginning: string, hoursOfInitialStage: number, proportionOfTraffic: number,
```
pedro823: If you're passing the context down, you don't need to pass the account since the context has it.

node/abTest/updateParameters.ts
```diff
@@ -13,22 +11,23 @@ import { InitializeParameters } from './initializeParameters'
...
+    workspacesData: WorkspaceData[], testingWorkspaces: TestingWorkspaces, testId: string): Promise<void> {
+        const { clients: { abTestRouter, storedash } } = ctx
```
pedro823: is it just me or is this indentation bigger than the others?

node/abTest/controller.ts
```diff
@@ -59,6 +59,6 @@ export const finishAbTestForWorkspace = async (ctx: ColossusContext) => {
   ctx.body = 'A/B Test finished'
 }
 
-export const updateParameters = async (ctx: EventsContext) => {
+export const updateParameters = async (ctx: Context) => {
   await update(ctx)
```
pedro823: This controller could just return the value.

node/abTest/controller.ts
```diff
@@ -14,7 +14,7 @@ export const timeToCompleteAbTest = async (ctx: ColossusContext) => {
   ctx.body = time
```
pedro823: Just something to consider in the future: these controllers are all returning texts. It would be nice if it returned JSON objects so that they're expandable with backwards compatibility (as adding a new key into an object doesn't break clients that do not care about that key).

node/utils/getWithRetries.ts
```diff
@@ -1,4 +1,4 @@
-export const getWithRetriesHelper = (handler: any) => async (retries: number, ctx: ColossusContext) => {
+export const getWithRetriesHelper = (handler: any) => async (retries: number, ctx: Context) => {
```
pedro823: Just a question, this functions logs to console. Don't you want it to log with `ctx.clients.logger` instead?

node/abTest/commands/initialize.ts
tergol1: Yes! Great!

node/utils/getWithRetries.ts
tergol1: Thanks for spotting this! Yes, I want!

node/abTest/controller.ts
tergol1: I couldn't agree with you anymore! I think that's a good idea for a next evolution.

PR #27, opened 2019-12-05
Feat/rev testingparams class

lucas-daher:
create a new class that extends testing parameters interface to be used when ab testing with revenue.

creating a new feature

just test if this didn't break anything, as the change isn't implemented yet.

#### Types of changes


node/typings/testingParameters.ts
```diff
@@ -67,6 +67,102 @@ export class TestingParametersConversion implements ITestingParameters{
...
+    constructor(testingWorkspaces: ABTestWorkspace[]) {
+        const parameters = testingWorkspaces !== null ? MapInitialParameters(testingWorkspaces) : new Map()
+        this.parameters = new Map(parameters)
+    }
+
+    public Get = (): Map<string, ABTestParameters> => {
+        return this.parameters
+    }
+
+    public Add = (workspaceName: string, abTestParameter = InitialABTestParameters) => {
```
tergol1: **nit** Just a doubt (maybe @pedro823 could answer that): Should we declare the type of `abTestParameter` as `ABTestParameters` like `abTestParameters: ABTestParameters = InitialABTestParameters `?

node/typings/testingParameters.ts
```diff
@@ -67,6 +67,102 @@ export class TestingParametersConversion implements ITestingParameters{
...
+            let avg = (2 * i + countTotal + 1) /2.0
+            U += avg * countIdx
+            i = j
+        }
+        U -= OrdersValueVector[idx].length * (OrdersValueVector[idx].length + 1) / 2.0
+        return U
+    }
+}
+
+interface aux {
```
tergol1: I think we should give this a better name for better readability.

node/typings/testingParameters.ts
```diff
@@ -67,6 +67,102 @@ export class TestingParametersConversion implements ITestingParameters{
...
+                countTotal++
+                j++
+            }
+            let avg = (2 * i + countTotal + 1) /2.0
+            U += avg * countIdx
+            i = j
+        }
+        U -= OrdersValueVector[idx].length * (OrdersValueVector[idx].length + 1) / 2.0
+        return U
+    }
```
tergol1: I love names with one letter but I must say that this makes readability worse haha
I would try to pick a name that says explicitly what is this method doing.

node/typings/testingParameters.ts
```diff
@@ -67,6 +67,102 @@ export class TestingParametersConversion implements ITestingParameters{
...
+
+    public Remove = (workspaceName: string) => {
+        this.parameters.delete(workspaceName)
+    }
+
+    public Update(workspacesData: Map<string, WorkspaceData>) {
+        const workspaces: aux = {workspace: [], OrdersValueVector: [], U: []}
+
+        for (const workspace of this.parameters.keys()) {
```
tergol1: **nit** I know that maybe I did this in the past but I think that `workspaceName` would be a better name for this variable as we could use workspace as a name for any class that contains properties of a workspace which contains, in particular, its name.

tergol1: Something that called my attention is the following line 
```
workspaces.workspace.push(workspace)
```

It's just harder to understand than it should be.

node/typings/testingParameters.ts
tergol1: But I really liked the idea of create a struct to help the iteration that you used this!

node/typings/testingParameters.ts
```diff
@@ -67,6 +67,102 @@ export class TestingParametersConversion implements ITestingParameters{
...
+                workspaces.OrdersValueVector.push(workspacesData.get(workspace)!.OrdersValueVector!)
+            }
+        }
+        
+        const size = workspaces.workspace.length
+        let sum = 0
+        for (let i = 0; i < size; i++) {
+            let U = this.U(workspaces.OrdersValueVector, i)
+            workspaces.U.push(U)
+            sum += U
```
tergol1: I think that this shadowing naming also made readability worse. Maybe we should think about a specific name for each instance of `U` - which I also prefer to has its name changed, as I'll say later.

node/typings/testingParameters.ts
tergol1: I would also ask for a comment clarifying what is being done here - why is this being calculated?

node/typings/testingParameters.ts
```diff
@@ -67,6 +67,102 @@ export class TestingParametersConversion implements ITestingParameters{
...
+        }
+        for (let i = 0; i < size; i++) {
+            this.parameters.set(workspaces.workspace[i], {a:(workspaces.U[i]/sum), b:0})
+        }
+    }
+
+    public UpdateWithFixedParameters = (proportion: number) => {
+        const actualProportion = proportion >= 0 && proportion <= 10000 ? proportion : 10000
```
pedro823: sounds like an auxiliar function 👀

node/typings/testingParameters.ts
```diff
@@ -67,6 +67,102 @@ export class TestingParametersConversion implements ITestingParameters{
...
+        for (const value of OrdersValueVector[idx]) {
+            v.push([value, 0])
+        }
+        for (let i = 0; i < OrdersValueVector.length; i++) {
+            if (i != idx) {
+                for (const value of OrdersValueVector[i]) {
+                    v.push([value, 1])
+                }
+            }
+        }
```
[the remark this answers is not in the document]
pedro823: the dude is good

node/typings/testingParameters.ts
```diff
@@ -78,6 +174,9 @@ export const createTestingParameters = (testType: TestType, testingWorkspaces: A
     if (testType == TestType.conversion) {
         return new TestingParametersConversion(testingWorkspaces)
     }
+    if (testType == TestType.revenue) {
```
pedro823: **nit**: I know this one's not your fault, but: always use `===` instead of `==`

node/typings/testingParameters.ts
```diff
@@ -67,6 +67,102 @@ export class TestingParametersConversion implements ITestingParameters{
...
+        const size = this.parameters.size
+        const nonMasterParameter = (10000 - actualProportion) / (size - 1)
+
+        for (const workspace of this.parameters.keys()) {
+            const parameter = workspace === MasterWorkspaceName ? actualProportion : nonMasterParameter
+            this.parameters.set(workspace, { a: parameter, b: 1 })
+        }
+    }
+
+    private U(OrdersValueVector: number[][], idx: number) {
```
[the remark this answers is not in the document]
tergol1: https://en.wikipedia.org/wiki/Mann%E2%80%93Whitney_U_test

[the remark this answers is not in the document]
pedro823: that's very _doido_

node/typings/testingParameters.ts
```diff
@@ -67,6 +67,102 @@ export class TestingParametersConversion implements ITestingParameters{
...
+        for (const workspace of this.parameters.keys()) {
+            if (workspacesData.has(workspace)) {
+                workspaces.workspace.push(workspace)
+                workspaces.OrdersValueVector.push(workspacesData.get(workspace)!.OrdersValueVector!)
+            }
+        }
+        
+        const size = workspaces.workspace.length
+        let sum = 0
+        for (let i = 0; i < size; i++) {
```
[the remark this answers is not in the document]
lucas-daher: because if I do this, it complains on the next line

node/typings/testingParameters.ts
lucas-daher: What? The whole `UpdateWithFixedParameters`? I didn't really get this comment

node/typings/testingParameters.ts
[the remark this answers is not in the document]
lucas-daher: haha it isn't, I'm just marking all the data with a 0 or 1 label for the test.

node/typings/testingParameters.ts
pedro823: no, the value seems like it could've been a function like [this one from router](https://github.com/vtex/kube-router/blob/master/src/Vtex.Kube.Router/PrivateAssemblies/Vtex.Kube.Router/Kubernetes/ComputingLimitsRanges.cs#L21), however I may be overengineering.

PR #28, opened 2019-12-10
Feat/storedash rev queries

lucas-daher:
Get detailed revenue data from storedash to use when AB testing with revenue as main metric.

Creating a new feature.

Linking the app, starting an ab test manually, checking that it fills `OrdersValueHistory`.

#### Types of changes


node/abTest/analysis/time/timeToComplete.ts
```diff
@@ -9,7 +9,9 @@ export async function TimeToCompleteAbTest(probability: number, storedash: Store
 }
 
 export async function AverageDaySessions(storedash: Storedash): Promise<number> {
-    const workspacesData = await storedash.getWorkspacesData('now-7d')
+    const beginning = new Date()
+    beginning.setDate(beginning.getDate() - 7)
+    const workspacesData = await storedash.getWorkspacesData(beginning.toISOString())
```
tergol1: Why do you want this change?

lucas-daher: This was a small refactor I did: `storedash.getWorkspacesData` now receives an ISO date string and makes the conversion to the query format itself. It needs to be this way when using revenue, and I think it's better to unify it like this.

node/clients/storedash.ts
```diff
@@ -26,12 +32,49 @@ export default class Storedash extends ExternalClient {
         }
         return workspacesData
     }
+
+    public getWorkspacesCompleteData = async (beginning: string): Promise<WorkspaceData[]> => {
```
tergol1: I just thinking about a possible name that explicitly says that it contains some kind of time series of order values.

PR #29, opened 2019-12-11
Feat/cache granular data

lucas-daher:
Create and use vbase as cache for the data collected from storedash queries.
When getting "granular data" we need cache, otherwise we would make too many queries (>4000 by the end of a 2 week AB test) too frequently (every ~5min).

Link the app, start an ab test, test the routes, check `workspaceDataCache.json`, see that everything works like a charm.

#### Types of changes


node/clients/storedash.ts
```diff
@@ -33,7 +33,7 @@ export default class Storedash extends ExternalClient {
         return workspacesData
     }
 
-    public getWorkspacesGranularData = async (beginning: string): Promise<WorkspaceData[]> => {
+    public getWorkspacesGranularData = async (beginning: string): Promise<{data: WorkspaceData[], updateTime: string}> => {
```
tergol1: WDYT about a type for this `{data: WorkspaceData[], updateTime: string}`? I really want to see your opinion because I'm not sure we should create it as we only use this one time.

node/clients/vbase.ts
```diff
@@ -35,6 +46,14 @@ export default class VBase extends BaseClient {
 
   public initializeABtest = async (initialTime: number, proportion: number, ctx: Context) => {
     const beginning = new Date().toISOString().substr(0, 16)
+    try {
+      const date = new Date()
+      const initialCache: WorkspaceDataCache = { ordersValue: [], lastUpdate: date.toISOString().substr(0, 16) }
```
tergol1: **nit** I think we should move this to the beginning of the document as it looks like a universal constant.

node/clients/storedash.ts
lucas-daher: I would rather not create a type as this is just a work around to return two values, the object is broken down into its values as soon as the function returns it.

node/clients/vbase.ts
lucas-daher: this isn't really constant as `date` is the current time the code is executed. We could have it as a global variable and update `date` every time it is called.

tergol1: Fixing the sentence before, "constant".  In the beginning 
```
const InitialWorkspaceDataCache = (date: Date) => { 
ordersValue: [], 
lastUpdate: date.toISOString().substr(0, 16)
}
```

and then use it as 

```
const initialCache = InitialWorkspaceDataCache(date)
```

lucas-daher: Great solution! Will implement

PR #31, opened 2020-01-21
Use try/catch better

tergol1:
Add "try/catches" for better error handling - adds a treatment to each possible error in a call to external clients - and fix the return of the function `TestWorkspaces`.

#### Types of changes


node/abTest/commands/finish.ts
```diff
@@ -7,22 +7,66 @@ import { TestWorkspaces } from '../testWorkspaces'
...
   const workspaceName = firstOrDefault(finishingWorkspace)
+  let workspaceMetadata: ABTestWorkspacesMetadata
+  let testData: VBaseABTestData
+  let testResults : TestResult[]
   try {
-    const workspaceMetadata = await abTestRouter.getWorkspaces(account)
-    const testingWorkspaces = new TestingWorkspaces(workspaceMetadata)
-    testingWorkspaces.Remove(workspaceName)
-    if (testingWorkspaces.Length() <= 1) {
+    workspaceMetadata = await abTestRouter.getWorkspaces(account)
```
athoscouto: Instead of wrapping promises on `try/catch` blocks you can chain them with `catch` or `then` functions.
Instead of:
```typescript
try {
    workspaceMetadata = await abTestRouter.getWorkspaces(account)
  } catch (err) {
    if (err.status === 404) {
      err.message = 'Test not found'
    }
    logger.error({ status: ctx.status, message: err.message })
    throw new Error(err)
  }
```
You can have:
```typescript
const logError = (err) => {    
    if (err.status === 404) {
      err.message = 'Test not found'
    }
    logger.error({ status: ctx.status, message: err.message })
    throw new Error(err)
  }
}
...
workspaceMetadata = await abTestRouter.getWorkspaces(account).catch(logError)
```

Since you can define the error handling logic elsewhere, the code becomes a lot clearer.

tergol1: Hum very nice! I'll change to do something like you suggested!

victorges: Pretty cool. Ends up pretty similar to what `go 2` cleaner error handling will look like 👀

PR #32, opened 2020-01-22
Fix VBase client

tergol1:
Fix VBase client that was trying to access a property of an undefined object.

This PR doesn't solve the root cause of this problem reported in the purpose part of this description but makes the experience doing A/B tests better to web-framework-team as they are getting lot of errors related to this. The root cause is related to the problem that the array of `TestResult`s shouldn't be empty in first place. Nevertheless this fix is necessary because we should treat correctly and don't explode everything when a non-predicted error occurs when this doesn't affect the completion of a determined task.

¯\_(ツ)_/¯ as it depends on a not tracked yet state to enter in the control part of the code that is altered by this PR.

#### Types of changes


node/clients/vbase.ts
```diff
@@ -85,14 +85,21 @@ export default class VBase extends BaseClient {
 
   public finishABtest = async (ctx: Context, results: TestResult[]): Promise<void> => {
     const testHistory = await this.fetchTestHistory(ctx)
-    if (!(results.length > 0) || results[0].ABTestBeginning !== testHistory.onGoing) {
+    let hasTestResult = true
+    if (!(results.length > 0)) {
```
athoscouto: ⛏ `results.length == 0` is easier to read IMO

PR #33, opened 2020-03-25
Fix/commands/update

caroldesenne:
Remove duplicated call to storage.getTestData
Stop creating a new error when we receive it from a called function (we throw away the error stack, otherwise)

#### Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)


node/abTest/commands/update.ts
```diff
@@ -17,13 +17,13 @@ export async function UpdateStatusOnEvent(ctx: Context): Promise<void> {
         hours = 0
         proportion = 100
       }
-      const testType = (await storage.getTestData(ctx)).testType
+      const testType = (testData).testType
```
victorges: **nit:** I think you don't need the parenthesis here anymore

PR #37, opened 2020-05-08
Fix error returned by calls to AB-Tester when no test is being performed

tergol1:
Fix error returned when there wasn't any test being performed.

We weren't returning early when we get the info that an account has none A/B test being performed and then called to VBase trying to get test's metadata, which didn't exist and then we return an 500 error.

All commands shouldn't change its behavior but finish and status, which should now return a 404 error when called in an account which has no A/B test being performed.

#### Types of changes


tergol1:
I don't know what happened but the commit's order is completely :kaotico: . Just to things make sense, I recommend to start the review by the commit with message
> Makes getWorksapces returns a TestingWorkspaces object

node/abTest/commands/finish.ts
```diff
@@ -7,18 +8,19 @@ import { TestWorkspaces } from '../testWorkspaces'
...
   } catch (err) {
-    if (err.status === 404) {
-      err.message = 'Test not found'
-    }
+    err.message = 'Error getting test metadata from router API'
     ctx.vtex.logger.error({ status: ctx.status, message: err.message })
     throw err
   }
+  if (testingWorkspaces.Length() === 0) {
+    ctx.response.status = 404
```
victorges, line 20: I think you don't need to set the `404` status code, the `NotFoundError` already carries that information and the root exception handler will set the corresponding status code for you.

node/abTest/commands/finish.ts
```diff
@@ -7,18 +8,19 @@ import { TestWorkspaces } from '../testWorkspaces'
...
   try {
-    workspaceMetadata = await abTestRouter.getWorkspaces(account)
+    testingWorkspaces = await abTestRouter.getWorkspaces(account)
   } catch (err) {
-    if (err.status === 404) {
-      err.message = 'Test not found'
-    }
+    err.message = 'Error getting test metadata from router API'
     ctx.vtex.logger.error({ status: ctx.status, message: err.message })
     throw err
```
victorges, line 17: I'm assuming this `err` will never be a `404` here, right? If so, we might want not to re-throw it.

node/abTest/commands/status.ts
```diff
@@ -1,12 +1,13 @@
...
 export async function ABTestStatus(ctx: Context): Promise<TestResult[]> {
   const { vtex: { account }, clients: { abTestRouter, logger, storage } } = ctx
   try {
-    const workspaces = await abTestRouter.getWorkspaces(account)
-    if (workspaces === null) {
-      ctx.response.status = 200
-      return []
+    const testingWorkspaces = await abTestRouter.getWorkspaces(account)
+    if (testingWorkspaces.Length() === 0) {
+      ctx.response.status = 404
```
victorges, line 9: :ditto:

node/clients/router.ts
```diff
@@ -11,8 +12,9 @@ export default class Router extends InfraClient {
         super('router', ctx, options, true)
     }
 
-    public getWorkspaces = async (account: string): Promise<ABTestWorkspacesMetadata> => {
-        return this.http.get<ABTestWorkspacesMetadata>(routes.Workspaces(account), { metric: 'abtest-get' })
+    public getWorkspaces = async (account: string): Promise<TestingWorkspaces> => {
+        const workspaceMetadata = await this.http.get<ABTestWorkspacesMetadata>(routes.Workspaces(account), { metric: 'abtest-get' })
```
victorges, line 16: Will this `get` throw an exception if some error is returned by the server? Like throwing an error corresponding to the received 404 here (that one with `err.response.status == 404` or sth like that).

If so, I believe you might need a `try/catch` here to return an empty `TestingWorkspaces` if a 404 is received (or re-throw the err if not a 404)

node/abTest/commands/finish.ts
tergol1: I thought that too but surprisingly `NotFoundError` don't do that and so seems necessary to set the status manually.

PR #40, opened 2020-07-30
Don't allow test to be initiated with NaN parameters

tergol1:
Don't allow A/B tests be initialized with NaN parameters. This don't allow AB-Tester to manage the weights of a test.

AB-Tester wasn't able to handle correctly the configuration of initial weights in a test due to `NaN` value saved for initial parameters.

Try to initiate a test with parameters (initial constraint of weights and time respecting this constraint) that are not numbers

#### Types of changes


node/abTest/commands/initialize.ts
```diff
@@ -9,9 +9,18 @@ export function InitializeAbTestForWorkspace(ctx: Context): Promise<void> {
...
-    return InitializeAbTest(Number(hoursOfInitialStage), Number(proportionOfTraffic), ctx)
+    const hoursOfInitialStageStr = firstOrDefault(hours)
+    const proportionOfTrafficStr = firstOrDefault(proportion)
+
+    const hoursOfInitialStage = Number(hoursOfInitialStageStr)
+    const proportionOfTraffic = Number(proportionOfTrafficStr)
+
+    if (Number.isNaN(hoursOfInitialStage) || Number.isNaN(proportionOfTraffic))
+    {
+        throw new Error(`An error occurred when reading values defining constants for proportion of traffic or the time respecting this initial constraint!`)
```
athoscouto, line 20: We should as specific as we can here. 
What about saying that we weren't able to parse he `hours` and/or `proportion` parameters as numbers?

tergol1: Nice! Very good suggestion!

PR #54, opened 2020-09-16
Replace logger

valentino-amadeus:
Replace `clients.logger` with `vtex.logger`. 

`clients.logger` is deprecated.

#### Types of changes


tergol1 (changes requested):
Everything is great with these changes, but I think you forgot to remove the file `clients/logger`.

valentino-amadeus:
There is no such file...

tergol1 (approved):
You're right! This client came from the `node-vtex-api`.

PR #55, opened 2020-09-17
Add option of selecting test type

valentino-amadeus:
Make it possible for the user to select the type of A/B test (either `conversion` or `revenue`) to be initialized.

If no type is selected, the test starts as a `conversion` test (the already existing users select no type on their requests and expect the test to be initialized as of type `conversion`). 

If the type is misspelled, an error is thrown. 

One should call the app, try to
1. start a test without specifying its type;
1. start a test specifying a (correctly spelled) type;
1. start a test setting a misspelled type

and verify whether the  event goes as expected.

#### Types of changes

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -13,7 +13,7 @@
     "@vtex/api": "^3.66.2",
     "tslint": "^5.20.1",
     "tslint-config-vtex": "^2.0.0",
-    "typescript": "3.7.3",
+    "typescript": "3.8.3",
     "typescript-map": "^0.0.7"
   },
   "scripts": {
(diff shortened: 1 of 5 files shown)


node/package.json
```diff
@@ -13,7 +13,7 @@
     "@vtex/api": "^3.66.2",
     "tslint": "^5.20.1",
     "tslint-config-vtex": "^2.0.0",
-    "typescript": "3.7.3",
+    "typescript": "3.8.3",
```
tergol1, line 16: Just to point something that I prefer when reviewing a PR but maybe it's just me and you shouldn't follow: these kind of stuff I like to almost ignore, then I prefer when these changes are in a separate commit.

tergol1 (approved):
LGTM!
Long have I waited for this moment...

node/package.json
valentino-amadeus: Oh, the "me" from the present absolutely agrees with you!

PR #57, opened 2020-09-18
Make the code look beautiful Part III

valentino-amadeus:
Reallocate `TestType` from `node/clients/vbase.ts` to `node/typings.d.ts`. 

`TestType` is a crucial concept throughout the app, so that it makes more sense to keep it in `typings.d.ts`.

No test needed. Once it passes the builder check, it shall work fine. 

#### Types of changes

### DIFF
--- a/node/clients/vbase.ts
+++ b/node/clients/vbase.ts
@@ -164,11 +164,6 @@ export default class VBase extends BaseClient {
   }
 }
 
-export enum TestType {
-  conversion,
-  revenue,
-}
-
 declare global {
   interface VBaseABTestData {
     dateOfBeginning: string
--- a/node/abTest/commands/initialize.ts
+++ b/node/abTest/commands/initialize.ts
@@ -1,5 +1,4 @@
 import { TSMap } from 'typescript-map'
-import { TestType } from '../../clients/vbase'
 import { createTestingParameters } from '../../typings/testingParameters'
 import { firstOrDefault } from '../../utils/firstOrDefault'
 import getRequestParams from '../../utils/BodyParser/getRequestParams'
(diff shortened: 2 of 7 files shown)


node/clients/vbase.ts
```diff
@@ -164,11 +164,6 @@ export default class VBase extends BaseClient {
   }
 }
 
-export enum TestType {
-  conversion,
-  revenue,
-}
-
```
tergol1, line 171: I think that remove this enum is what is bothering me about this change. Could you explain a bit your intention here?

tergol1 (changes requested):

node/clients/vbase.ts
valentino-amadeus: Removing the enum is all the PR does 😅 (together with adding it as a type to `typings.d.ts`, of course).
It's an improvement on the code's logic. The test type is a crucial concept throughout the code, so it doesn't make sense to define it in such a singular file and to import it everywhere else. `typings.d.ts` is the perfect place to insert foundational concepts.

tergol1: Isn't it possible to make the global `TestType` a enum? I don't like to have always to write the exact sequence of characters everywhere.

valentino-amadeus: I'm with you on that, but unfortunately you can't define enums in `typings.d.ts`, as they are constants - not types.

valentino-amadeus: And even though we don't like the idea of writing the whole words every time, it does have an up side: we're not gonna confuse the types. When dealing with enums, we only specify the index, so that may bring confusion - or at least the obligation of checking what type has each index, which already consumes the time we'd spend writing the words 😅 .

tergol1 (approved):

PR #71, opened 2020-09-24
Fix bugs in commands/update

valentino-amadeus:
- Stop using default parameters - when updating - only because the initial time was set to 0;
- Set the default proportion to 10000.

When the test is updated, the app checks if there is any problem with the parameters and, if so, sets them to the default parameters values, to proceed the updating. So far, this resetting has been being executed no matter what when the initial time is 0, and this is not what we want: the initial time being 0 is not a problem at all. 
Also, the default value of the `proportion` parameter is `10000`.

#### Types of changes


tergol1 (approved):
LGTM
There he is! You had already fixed the bug we are facing in Carrefour's case. Very nice!

PR #75, opened 2020-09-25
Fix GetGranularData

valentino-amadeus:
Make the `CalculateOrdersValue` function return either an empty array, in case no order was made or no money was spent, or  an array containing as many elements as the number of sessions, each element being equal to the average of money spent per session in the sessions under consideration.

We want to have in hands - when doing the analysis - the set of orders' values (each value corresponding to one order and vice-versa), with no zero included. 
In the way `CalculateOrdersValue` worked, zero values might have been appearing, and orders might have been left aside: no matter how many orders we had in each `5` minutes window, the function returned one and exactly one value. 

#### Types of changes

### DIFF
--- a/node/clients/storedash.ts
+++ b/node/clients/storedash.ts
@@ -68,7 +68,7 @@ export default class Storedash extends ExternalClient {
                     const m: StoreDashResponse = metrics[i] as unknown as StoreDashResponse
                     const workspaceData = workspacesData.get(m['workspace'])
                     if (workspaceData !== undefined) {
-                        workspaceData.OrdersValueHistory.push(CalculateOrdersValue(m['data.ordersValue'], m['data.sessionsOrdered']))
+                        workspaceData.OrdersValueHistory.push(...CalculateOrdersValue(m['data.ordersValue'], m['data.sessionsOrdered']))
                     }
                 }
                 dateFrom.setTime(dateTo.getTime())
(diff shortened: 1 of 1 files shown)


node/clients/storedash.ts
```diff
@@ -82,10 +82,17 @@ export default class Storedash extends ExternalClient {
...
+    let ordersValue: number[] = []
+
+    if (value <= 0 || sessions <= 0) {
+        return ordersValue
+    }
+    const avarageValue = value/sessions
+    for (let i = 0; i < Math.round(sessions); i++) {
+        ordersValue.push(avarageValue)
+    }
+    return ordersValue
```
tergol1, line 95: Shouldn't we check if these values are numbers? If so, I think that the `sessions === NaN` and `value === NaN` should be in that line.

valentino-amadeus: `m` is of type `StoreDashResponse`, so `m['data.ordersValue']` and `m['data.sessionsOrdered']` are both of type `number` (check line 71, where `CalculateOrdersValue` is called).

valentino-amadeus: Oh, I didn't know `NaN` was of type `number` \o/. 
Gonna fix that.

valentino-amadeus: I made it return a standard result, as it did before. But I've been thinking... Shouldn't we throw an error?

tergol1 (approved):

node/clients/storedash.ts
tergol1: I think the problem here to question if a value is a number is due to too big/small numbers that could lead to a `NaN`. So I think is better to maintain a default value instead of returning an error that could break a test being executed. But I'm not entirely convinced. I think we can look into that decision in the future

PR #76, opened 2020-09-29
Check workspace name before initializing and finishing

valentino-amadeus:
- Avoid initializing a test with a non-existing workspace, with a development workspace or with the `master` workspace. 
- Throw an error when trying to remove (`finish`) from the test the master workspace or a workspace that was not under test.

- When initializing a test with a non-existing or with the master workspace, problems would appear at the Router.
- When removing from the test a workspace that was not being tested (that could occur because of a misspelling, for example), the set of workspaces under test would remain the same, but the proportions would be reset. 

- Call the initializing routes, set as workspace: 
    1. a name that does not correspond to any of the account's workspaces;
    2. a development workspace;
    3.  the `master` workspace;
    4. a production workspace
and verify whether the app responds as expected*.
- Call the finishing route, set as workspace:
    1. a name that does not correspond to any of the account's workspaces;
    2. one of the account's workspaces that are not under test;
    3. the `master` workspace;
    4. one of the workspaces under test
and verify whether the app responds as expected*.

\* <sub>You should also call the test's `status` to check if everything went as intended.</sub> 

#### Types of changes

### DIFF
--- a/node/abTest/commands/initialize.ts
+++ b/node/abTest/commands/initialize.ts
@@ -5,39 +5,36 @@ import getRequestParams from '../../utils/BodyParser/getRequestParams'
 import { concatErrorMessages } from '../../utils/errorHandling'
 import { WorkspaceMetadata } from '@vtex/api'
 
-export async function InitializeAbTestForWorkspace(ctx: Context): Promise<void> {
+export function InitializeAbTestForWorkspace(ctx: Context): Promise<void> {
     const workspace = ctx.vtex.route.params.initializingWorkspace
-    const workspaceName = firstOrDefault(workspace)
-    await CheckWorkspace(workspaceName, ctx)
-
-    return InitializeAbTest(workspaceName, 1, 5000, ctx)
+    return RunChecksAndInitialize(ctx, workspace, '1', '5000', 'conversion')
 }
 
-export async function InitializeAbTestForWorkspaceWithParameters(ctx: Context): Promise<void> {
-    const { vtex: { route: { params: { hours, proportion, initializingWorkspace } } }} = ctx
-    const [ workspaceName, hoursOfInitialStage, proportionOfTraffic ] = [ initializingWorkspace, hours, proportion ].map(firstOrDefault)
-    await CheckWorkspace(workspaceName, ctx)
-    
-    checkIfNaN(hoursOfInitialStage, proportionOfTraffic)
-    const [ numberHours, numberProportion] = [ Number(hoursOfInitialStage), CheckProportion(Number(proportionOfTraffic))]
-    
-    return InitializeAbTest(workspaceName, numberHours, numberProportion, ctx)
+export function InitializeAbTestForWorkspaceWithParameters(ctx: Context): Promise<void> {
+    const { vtex: { route: { params: { hours, proportion, initializingWorkspace } } }} = ctx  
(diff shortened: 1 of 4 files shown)


node/abTest/commands/initialize.ts
```diff
@@ -82,4 +88,24 @@ const checkIfNaN = (hours: string, proportion: string) => {
 
 const CheckProportion = (proportion: number): number => {
     return proportion >= 0 && proportion <= 10000 ? Math.round(proportion) : 10000
+}
+
+const CheckWorkspace = async (workspaceName: string, ctx: Context) => {    
+    if (workspaceName === 'master') {
+        throw new Error(`Bad workspace name: please select another workspace`)
```
tergol1, line 95: I think you should add the information about why this was not allowed.

node/abTest/commands/initialize.ts
```diff
@@ -2,17 +2,21 @@ import { TSMap } from 'typescript-map'
...
 import { firstOrDefault } from '../../utils/firstOrDefault'
 import getRequestParams from '../../utils/BodyParser/getRequestParams'
+import { concatErrorMessages } from '../../utils/errorHandling'
+import { WorkspaceMetadata } from '@vtex/api'
 
-export function InitializeAbTestForWorkspace(ctx: Context): Promise<void> {
+export async function InitializeAbTestForWorkspace(ctx: Context): Promise<void> {
     const workspace = ctx.vtex.route.params.initializingWorkspace
     const workspaceName = firstOrDefault(workspace)
+    await CheckWorkspace(workspaceName, ctx)
```
tergol1, line 11: Looking this repetition that you introduced in this code I realized that we were already repeating a lot of things here. WDYT of creating a helper `initializeAbTestForWorkspace ` that has as parameters everything that could be used to initialize a test, i.e. workspace names, parameters, etc, and then we use only small functions - the ones that we have today, but simpler - to only parse these parameters and then call this helper with what is necessary?

tergol1 (changes requested):
We're almost there! I just proposed something to avoid some code repetition, but you can implement it the way you prefer. Then, I'll take a look again.

node/abTest/commands/initialize.ts
valentino-amadeus: Right! I thought the message would become too large, but since you find that that would be better too, I'll do it!

node/abTest/commands/initialize.ts
valentino-amadeus: That seems awesome! This repetition has been bothering me for a while haha.

node/abTest/commands/initialize.ts
```diff
@@ -71,6 +74,12 @@ async function InitializeWorkspaces(ctx: Context, id: string, testingWorkspaces:
     })
 }
 
+const checkTestType = (Type: string) => {
+    if (Type !== 'conversion' && Type !== 'revenue') {
```
tergol1, line 78: WDYT about creating constants for these strings? We'll probably use it more times in the code.

tergol1 (approved):
LGTM

node/abTest/commands/initialize.ts
valentino-amadeus: They were constants before, but I changed it. https://github.com/vtex/ab-tester/pull/57
It bothers me that they are not constants, but it bothers me even more if they are defined in a file other than `typings.d.ts`. Do you see any drawbacks in the way these types are currently defined?

PR #78, opened 2020-09-30
Improve error handling

valentino-amadeus:
Specify the error's status and return the error's message in the response's body.

Call all the routes involved setting both good and bad parameters and check whether the response (and the status) are formed as expected.

#### Types of changes

### DIFF
--- a/node/abTest/analysis/compareWorkspaces/revenue.ts
+++ b/node/abTest/analysis/compareWorkspaces/revenue.ts
@@ -13,7 +13,9 @@ export async function EvaluateRevenue(abTestBeginning: string, workspaceAData: W
     const ordersHistoryB = workspaceBData.SinceBeginning.OrdersValueHistory
 
     if (ordersHistoryA.length <= 10 || ordersHistoryB.length <= 10) {
-        throw new Error('Not enough data to perform accurate analysis')
+        const err = new Error('Not enough data to perform accurate analysis') as any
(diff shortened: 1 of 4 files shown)


node/abTest/analysis/compareWorkspaces/revenue.ts
```diff
@@ -13,7 +13,9 @@ export async function EvaluateRevenue(abTestBeginning: string, workspaceAData: W
     const ordersHistoryB = workspaceBData.SinceBeginning.OrdersValueHistory
 
     if (ordersHistoryA.length <= 10 || ordersHistoryB.length <= 10) {
-        throw new Error('Not enough data to perform accurate analysis')
+        const err = new Error('Not enough data to perform accurate analysis') as any
+        err.status = 404
```
valentino-amadeus, line 17: Not sure what status code would best fit here.

tergol1: Very good point. I think we should stop to return an error here and start to return early the info that there is not enough data. The status should be then a `2xx`.

tergol1 (approved):

node/abTest/analysis/compareWorkspaces/revenue.ts
valentino-amadeus: This error throw has been removed somewhere in the future. :nazare:

PR #79, opened 2020-09-30
Fix reinitialization

valentino-amadeus:
When a workspace is removed from the test (through the `finish` command) or when a random restart is carried out when updating, the test is reinitialized. This means that the collecting of data is restarted, with the remaining workspaces receiving (initially, during the amount of time selected by the user in their last `initialize` command) amounts of traffic in accordance with the proportion set by the user in the last initialization. In this PR, we fix two bugs:
- We make the test be reinitialized with the correct proportions of traffic - rather than with a homogeneous distribution, as it's been being done so far.
- We reset the date of beginning (in `VBase`) to the current time. With that, the app will only consider the data collected after the reinitialization when performing the test's result's calculations.

#### Types of changes


node/typings/testingParameters.ts
```diff
@@ -106,7 +106,7 @@ class TestingParametersRevenue implements ITestingParameters{
             sum += U
         }
         for (let i = 0; i < size; i++) {
-            this.parameters.set(testData.workspaceNames[i], { a: Math.round(testData.U[i]/sum), b: 0 })
+            this.parameters.set(testData.workspaceNames[i], { a: Math.round(10000*testData.U[i]/sum), b: 0 })
```
tergol1, line 110: More a meta review - and I think you already learnt that since you opened this PR: in cases like that, when you're fixing a bug and the reason why it was a bug or why the change fixes it is not clear, it's nice to write a comment for the commit explaining it.

node/clients/storedash.ts
```diff
@@ -82,10 +82,17 @@ export default class Storedash extends ExternalClient {
...
-    return (valueCleaned/sessionsCleaned)
+const CalculateOrdersValue = (value: number, sessions: number): number[] => {
+    let ordersValue: number[] = []
+
+    if (isNaN(value) || isNaN(sessions) || value <= 0 || sessions <= 0) {
+        return ordersValue
+    }
+    const avarageValue = value/sessions
+    for (let i = 0; i < Math.round(sessions); i++) {
+        ordersValue.push(avarageValue)
```
tergol1, line 93: I think there is a misspelling in `avarageValue`, which would be `averageValue`

tergol1 (approved):
LGTM

node/typings/testingParameters.ts
valentino-amadeus: This commit was in a PR of its own. I have no idea how it ended up here. :think-360:

node/clients/storedash.ts
valentino-amadeus: Nice. I found another instance of this typo. Gonna open a PR - :donotwant: to rebase everything.

node/typings/testingParameters.ts
valentino-amadeus: [Here it is](https://github.com/vtex/ab-tester/pull/73). whatisgoingon :wasabihelp:

PR #83, opened 2020-10-07
Return best workspace overall

valentino-amadeus:
When the test's `status` is called, return the best performing workspace over all the workspaces that took part in the test.

For now, the user only gets a list of comparisons between the challenger workspaces and the `master`. In case more than one of the challengers beat the master, there is no clear way of getting the information of which workspace should be promoted.

#### Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)


valentino-amadeus:
Looking back at this, I think we could use the comparisons of each workspace with the master to choose the winner. That would avoid extra calculations. What do you say, dear reviewer?

tergol1 (approved):

PR #84, opened 2020-10-08
Refactor commands/initialize and commands/finish

valentino-amadeus:
Gather functions that set initial parameters and workspaces in the `Router` together and use them to simplify the code of `initialize.ts` and `finish.ts`.

The code was unnecessarily confusing, as the functionalities of the already defined functions were being implemented manually. Now, the writing is cleaner and the functions' code rationale is much more easily understandable. 

#### Types of changes


valentino-amadeus:
Commit _start test with new id_ fixes a bug in `commands/initialize.ts` and `commands/finish.ts`: whenever starting a new test, we must set a new `test id`. So far, we have been reusing the `id` from the previous test.

node/abTest/initialize-Router.ts
```diff
@@ -12,4 +12,11 @@ export async function InitializeParameters(ctx: Context, id: string, testingWork
         Id: id,
         parameterPerWorkspace: tsmap,
     })
+}
+
+export async function InitializeWorkspaces(ctx: Context, id: string, testingWorkspaces: ABTestWorkspace[]): Promise<void> {
+    await ctx.clients.abTestRouter.setWorkspaces(ctx.vtex.account, {
+        id: (id),
+        workspaces: testingWorkspaces,
+    })
```
valentino-amadeus, line 21: Dear reviewer, WDYT about putting this into `clients/router.ts`?

valentino-amadeus: The whole file (i.e., both functions), I mean.

node/abTest/updateParameters.ts
```diff
@@ -6,7 +6,7 @@ import { FilteredWorkspacesData } from '../utils/workspace'
 import { MapWorkspaceData } from '../utils/workspacesInfo/workspaceData'
 import { IsInitialStage } from './analysis/time/initialStage'
 import { BuildCompleteData } from './data/buildData'
-import { ResetParameters } from './initializeParameters'
+import { InitializeParameters } from './initialize-Router'
```
tergol1, line 9: I think `initialize-Router` isn't a real good name for this file. First, we usually prefer camelCase instead of kebab-case. But my biggest problem was with the idea of making an abstract concept, to create initial values for the routing parameters, into something concrete (and the specialized) calling it something specific to do with a router (`kube-router`, in our case).

I think it would be better to maintain this name more abstract/generic, as ab-tester can be, in the future, the system that configures traffic division of other router (which can possibly don't be a router but something that contains the routing logic).

node/abTest/updateParameters.ts
```diff
@@ -40,7 +40,7 @@ export async function UpdateParameters(ctx: Context, aBTestBeginning: string, ho
             })
             return
         }
-        await InitializeParameters(ctx.vtex.account, testingWorkspaces.ToArray(), proportionOfTraffic, testType, testId, abTestRouter)
+        await InitializeParameters(ctx, testId, testingWorkspaces.ToArray(), proportionOfTraffic, testType)
```
tergol1, line 43: For consistence I think you should pick one of `testId` or `id` and only use it in this change.

node/abTest/initialize-Router.ts
tergol1: Sounds good!

tergol1: But regarding my other comment, I don't think you should delete this file as its function is more of a generic entity to configure any routing system able to split traffic between different versions.

tergol1 (approved):
I think it could be a good improvement for this code a change in the renaming of a file I mentioned in the comments, but overall LGTM and certainly it does what it should.

node/abTest/initialize-Router.ts
valentino-amadeus: Very good point!

node/abTest/updateParameters.ts
valentino-amadeus: I love the idea!

node/abTest/updateParameters.ts
valentino-amadeus: Where am I using `id`?

node/abTest/updateParameters.ts
valentino-amadeus: >First, we usually prefer camelCase instead of kebab-case

We do have (_i.e._, we already had) kebab-cased file names (`node/utils/evaluation-response.ts`, for instance). Just so that I get a good understanding of the idea: are you saying those were bad name choices?

tergol1: Yes! And I did this myself in the past hahaha

node/abTest/updateParameters.ts
tergol1: Ahhhh the commit tricky hahaha
Yeah, you're not using at all

node/abTest/updateParameters.ts
valentino-amadeus: Alright, it is good to know. I'll fix them all together.

valentino-amadeus: But in another PR, because of all of the rebasing this change would require. 
I'll take a note so that we do this when we reorganize the whole repo.

PR #95, opened 2020-11-30
Specify analysis approach when initializing and updating test

valentino-amadeus:
Enable user to choose the test's analysis approach (`frequentist` or `bayesian`). 
The appropriate responses are going to be implemented in a further PR (for the sake of readability, given the big changes the implementation is going to demand).

#### Types of changes


node/utils/Request/Checks.ts
```diff
@@ -2,7 +2,7 @@ import { concatErrorMessages } from '../../utils/errorHandling'
 import { WorkspaceMetadata } from '@vtex/api'
 import TestingWorkspaces from '../../typings/testingWorkspace'
 
-const expectedFields = ['InitializingWorkspaces', 'Hours', 'Proportion']
+const expectedFields = ['InitializingWorkspaces', 'Hours', 'Proportion', 'Type']
```
tergol1: What do you think about a qualifier to `Type`? Not sure, but thinking about something like `MetricType`.

node/typings/testingParameters.ts
```diff
@@ -4,20 +4,22 @@ import { InitialABTestParameters, WorkspaceToBetaDistribution } from '../utils/w
 
 const MasterWorkspaceName = 'master'
 
-interface ITestingParameters {
+interface TestingParameters {
```
tergol1, line 7: It would be nice to maintain the `I` in `ITestingParameters` (we can quicly identify that it's an intterface)? We can change the name here or change the other to differentiate between these two `TestingParameters`.

tergol1 (approved):
LGTM!
That PR was really well written!

node/typings/testingParameters.ts
valentino-amadeus: Yeah, the naming is not good... We'll think of something better when we refactor the repo as a whole.

node/utils/Request/Checks.ts
valentino-amadeus: I tried to keep the mood of the other fields' names - very short, even though not totally explanatory. I didn't like it much, but I thought it wouldn't be a problem, as there would be documentation. 
But we sure can use better descriptive names when we refactor the repo as a whole.

PR #96, opened 2020-12-01
Structure responses accordingly to test's type and approach

valentino-amadeus:
Create response forms, one for each of the four possible testing contexts, and manage to respond in accordance to them (when giving test's status and when giving test's result).

<sub>The already implemented revenue evaluation (Mann-Whitney test) is going to be replaced with a truly frequentist revenue analysis.</sub>

#### Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)


node/abTest/analysis/compareWorkspaces/frequentistRevenue.ts
```diff
@@ -1,60 +1,9 @@
...
-    const pValueRevenue = PValueRevenue(uValue, ordersHistoryA.length, ordersHistoryB.length)
-    const winnerRevenue = PickWinner(pValueRevenue, workspaceAData.SinceBeginning.Workspace, workspaceBData.SinceBeginning.Workspace)
-    const effectSizeA = CalculateEffectSize(uValue, ordersHistoryA.length, ordersHistoryB.length)
-    const effectSizeB = 1 - effectSizeA
-    const medianA = CalculateMedian(workspaceAData.SinceBeginning.OrdersValueHistory)
-    const medianB = CalculateMedian(workspaceBData.SinceBeginning.OrdersValueHistory)
-
-    return EvaluationResponseRevenue(abTestBeginning, workspaceAData, workspaceBData, winner, lossA, lossB, probabilityTwoBeatsOne, PValueConversion, 
-        winnerRevenue, pValueRevenue, effectSizeA, effectSizeB, medianA, medianB)
+export function Evaluate(abTestBeginning: string, workspaceA: WorkspaceCompleteData, workspaceB: WorkspaceCompleteData): FrequentistEvaluationResultRevenue {
```
tergol1, line 3: I didn't understand this change as well. You just removed the previous implementation but didn't add back anything, just a draft?
If that's the case, I don't think we should remove this code.

tergol1 (changes requested):
I think this change potentially broke ton of things related to ab testing so need some things to be different here so we can get it working as it is expected.

node/typings.d.ts
```diff
@@ -32,34 +32,80 @@ declare global {
     Winner: string
   }
 
-  interface EvaluationResult {
```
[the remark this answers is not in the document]
valentino-amadeus: Yes, it will break the contract, and that's why it is a breaking change.

node/utils/evaluation-response/default.ts
```diff
@@ -0,0 +1,88 @@
...
+    AverageRevenueA: NaN,
+    AverageRevenueB: NaN,
+    WorkspaceASessionsLast24Hours: 0,
+    WorkspaceBSessionsLast24Hours: 0,
+    AverageRevenueALast24Hours: NaN,
+    AverageRevenueBLast24Hours: NaN,
+    ProbabilityAbeatsB: NaN,
+    ProbabilityBbeatsA: NaN,
+    ExpectedLossChoosingA: NaN,
+    ExpectedLossChoosingB: NaN
```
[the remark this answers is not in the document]
valentino-amadeus: Maybe we should update the contract - we're gonna be doing that anyway, as this PR introduces breaking changes. I don't like returning `0` when we don't have enough information to perform the calculations, because `0` is a well-defined value (it could very well be the actual result). But looking back now I understand that we can inform the user about the situation in another response field.

node/abTest/analysis/compareWorkspaces/frequentistRevenue.ts
valentino-amadeus: Yes, this is a draft. The actual function is implemented in [#103](https://github.com/vtex/ab-tester/pull/103).

valentino-amadeus: The idea is that we will merge both PRs (and all the intermediate ones) at once.

PR #97, opened 2020-12-02
Fix bayesian analysis of conversion

valentino-amadeus:
Remove everything related with `P-Value` calculations in the bayesian analysis of conversion. 

Bayesian analyses make all the decisions based on `expected losses`. The concept of a `P-Value` does not make sense in the context of a bayesian analysis, once there is no uncertainty on the distributions we work with: all the probabilities we calculate during the analysis are unquestionably certain, given the priors we had and the events we observed.

#### Types of changes

### DIFF
--- a/node/utils/mathTools/decisionRule/bayesianConversion.ts
+++ b/node/utils/mathTools/decisionRule/bayesianConversion.ts
@@ -1,8 +1,5 @@
 import { WorkspaceToBetaDistribution } from '../../workspace'
 import { logBeta } from '../forBetaDistribution/beta-function'
-import { pValue } from '../forBetaDistribution/statistics/samplesRestrictions'
-
-const Confidence = 0.95
 
 /*
 *   The reason for using the function logBeta is that we're dealing with very large numbers and the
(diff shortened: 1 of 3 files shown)


node/utils/mathTools/decisionRule/bayesianConversion.ts
```diff
@@ -101,26 +98,14 @@ export function LossFunctionChoosingVariantOne(Beta1: ABTestParameters, Beta2: A
...
-        return WorkspaceA.Workspace
-    }
-    else if (chooseB && distinct) {
-        return WorkspaceB.Workspace
-    }
-    else {
-        return null
-    }
+    if (lossA < lossB) return WorkspaceA.Workspace
+    if (lossA > lossB) return WorkspaceB.Workspace
```
tergol1, line 109: Why did you want to remove the bound on `epsilon`? Don't you think it is nice to wait the loss become small to star to declare a workspace the winner of the test?

tergol1 (commented):
Before approve this PR I would like to discuss one point.

node/utils/mathTools/decisionRule/bayesianConversion.ts
valentino-amadeus: I figured that we would probably discuss that and that we may and up delivering to the user a `"Not decided yet"` result - and perhaps a `"draw"` even when the losses are not exactly the same - for the sake of the user's decision's safety, but the point is that at the time I thought that this did not fit in a bayesian analysis: a bayesian analysis delivers, at every moment, an "absolute" result; a computation of losses and probabilities that is undoubtedly correct given the previous and the observed data; so if a version of the store provides less losses when compared to the other, it should be the winner. I thought that it wouldn't be a big trouble since the user is going to receive the expected losses of each version, so that they is going to see if the losses are not small or if they are too close to each other. 
But I do understand the idea of saying that no version can be declared winner or that the two versions are equivalent (even though they aren't) - I just wanted to have this conversion ;P . What do you say?

tergol1: I think the point here is about a stopping point of an algorithm. The `epsilon` is not related to a bayesian analysis, but to a decision that the use wants to make. I think a better approach would be let the user chooses this value but, as it is a complicated concept, we should give him a default value.

tergol1: If you like this suggestion. I think it would be better to don't remove that parameter in this PR and then open a follow-up implementing the whole change: add this parameter to test's metadata and create an API to allow user to configure it - then we will always start the test with a default value.

valentino-amadeus: Nice suggestions. They make absolute sense. Depending on how we decide that the analysis is actually going to be, we may even create a new layer, that takes into consideration these "confidence" aspects.

PR #127, opened 2021-03-19
Atomicity

valentino-amadeus:
`ab-tester` saves data through both `Kube-Router` and `VBase`. As if that wasn't enough, it handles more than one file in each of them. As a result, each operation of initialization or finalization of a test includes several IO calls. And as in any scenario like that, it is possible that some of these calls succeed whereas others don't. It is thus very important for us to reflect on the atomicity of the ab testing operations. The purpose of this PR is:
1. to make test initialization atomic: we either initialize the test in both `Kube-Router` and `VBase`, or we don't initialize the test at all;
2. to declare that the test has been finished as soon as the traffic division stops: it will be no longer possible that `ab-tester` stops the traffic division but throws an error, stating that the test could not be finished;
3. to remove the hierarchy that existed between the finalization operations in `VBase` (saving results and deleting files): now, `ab-tester` will perform as many operations as it can, regardless of whether any of them results in error.

Some checks have been removed as well, and an important side effect of these removals is that **it is now possible to start a test with 0% of traffic going into the `master` workspace**.

#### Types of changes


node/abTest/commands/initialize.ts
```diff
@@ -64,4 +63,23 @@ async function InitializeAbTest(workspacesNames: string[], hoursOfInitialStage:
...
+
+    const { vtex: { account }, clients: { abTestRouter, storage } } = ctx
+    try {
+        await InitializeProportions(ctx, workspaces.Id(), workspaces.ToArray(), masterProportion)
+        await InitializeWorkspaces(ctx, workspaces.Id(), workspaces.ToArray())
+        await storage.initializeABtest(workspaces.WorkspacesNames(), hoursOfInitialStage, masterProportion, testType, testApproach, isMAB, ctx)
+    } catch (err) {
+        try {
+            await abTestRouter.deleteProportions(account)
+            await abTestRouter.deleteWorkspaces(account)
```
tergol1, line 79: Think about what happens when the later calls fails: We can be in any state given by some of the original calls being successful or not. So, most important than delete any data previously saved (which I will say something about it) we should guarantee that any intermediate state is consistent.

If we initialize the workspaces in kube-router and not in VBase, aab-tester will be in a potentially inconsistent state, as it won't have any metadata about the test and it won't know how to update parameters, which it would try as it has the information from kube-router/workspaces saying that some test is running. 

So to think about this problem I think we can discuss the hierarchy of data. Since the beginning we thought it as

- `workspaces` parameter saved in Metadata (Kube-Router)
- `parameters` parameter saved in Metadata (Kube-Router)
- ab test metadata parameter saved in VBase

This should be read as "If the data above is saved, then certainly the ones below are as well". So, considering this order of priorities, I think should use it to save or delete data.

Then, my first proposal is to save first the metadata on VBase, then the parameters on Metadata (Kube-Router) and finally the workspaces on Metadata.

But let me go a little bit further on this line. Following this order I proposed ab-tester is able to "recover itself" from errors during these IO calls in its logic. What I'm saying is: if the `workspaces` parameter isn't saved, the other data will be completely  ignored by ab-tester. So it will be a meaningless data saved - and very small one. But as someone tries again to initialize the test, we can make the operations of saving `parameters` and `abTestMetadata` overwrite anything that was saved before (if you prefer scenarios being completely descriptive by the code, we can check the `workspaces` parameter first and then, if it wasn't there, we can do this overwrite thing). If we do this we can even leave these two extra calls in case of some error when trying to initialize all the stuff.

**P.S.** sorry for the wall-text

tergol1 (changes requested):
I stopped in the beginning of the commit path butt I think that my proposal can lead to a lot of changes or, at least, to a good discussion.

node/abTest/commands/initialize.ts
valentino-amadeus: >P.S. sorry for the wall-text

No problem haha

valentino-amadeus: I realize that I have made a mistake. I thought that the metadata (that ab-tester saves in VBase) could be adjusted in the updates, in case ab-tester had problems with saving it in the first attempt. But, in our current architecture, we cannot afford to do that. In fact, it is possible that the metadata's deletion fails at the end of a test, and that would lead to serious trouble in the next test's updates. We will talk more about this - I have a proposal for a change in abtest's metadata's architecture -, but for now, I will adapt the tests' initialization to our current scenario. 

On the other hand, I do not know if I agree with your proposal. Doesn't Kube-Router start to split the traffic as soon as it receives the test's `proportions`? If we don't save the `workspaces` first, we'll have trouble when trying to update.

tergol1: Remember that if any call fails, then we will return an error. So the client will try to call again ab-tester and in the meanwhile we didn't tried to update the parameters as we don't have saved the `workspaces`.

valentino-amadeus: So Kube-Router starts the traffic split as soon as it receives the `proportions` but, at the same time, starts the updatings as soon as it receives the `workspaces`?

valentino-amadeus: Sounds like we should do something about it in `Kube-Router`, if that's the case.

tergol1: We can also do something there! Maybe we can change these two calls by only one and then handle in Kube-Router this logic of saving on Metadata both data.

I would suggest to create a new API on Kube-Router to do that thing and then alter AB-Tester to use it.

tergol1: But then we should try to save the files in the order 1) ab test metadata 2) traffic parameters on Metadata

valentino-amadeus: >But then we should try to save the files in the order 1) ab test metadata 2) traffic parameters on Metadata

Sure!

valentino-amadeus: >Maybe we can change these two calls by only one and then handle in Kube-Router this logic of saving on Metadata both data.

Great idea!

PR #94, opened 2020-11-25
Implement bayesian revenue math

valentino-amadeus:
#### What is the purpose of this pull request?
Implement mathematical analysis for the bayesian revenue test.

#### How should this be manually tested?
Call the functions from `decisionRule/bayesianRevenue.ts` with a bunch of random (positive integer) parameters and compare the results with what they should be - the functions from `decisionRule/bayesianConversion.ts` may be useful in the process.

#### Types of changes
- [x] Math stuff (non-breaking change which also doesn't add any functionality (yet))
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Requires change to documentation, which has been updated accordingly.

PR #133, opened 2021-04-01
Add unused math stuff that's been removed

valentino-amadeus:
As requested [here](https://github.com/vtex/ab-tester/pull/94#pullrequestreview-626631802).


node/utils/mathTools/unusedTools/betaDistribution.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    if (x <= 0) return 0
+    if (x >= 1) return 1
+
+    let value = 1 - Math.pow(1-x, b)
+    for (let i = 1; i < a; i++) {
+        value -= Math.exp( i*Math.log(x) + b*Math.log(1-x) - Math.log(i) - logBeta(i, b) )
+    }
+    return value
+}
```
valentino-amadeus, line 29: This is different from what we had before. Not only because it is the function's regularized version, but also because I've looped over the first parameter - so that, in our presumed use cases, the function is more performatic.

tergol1 (approved):
LGTM!

I think that my suggestion would be a good improvement for these functions (even tough I know we don't use them nowadays)

node/utils/mathTools/unusedTools/betaDistribution.ts
valentino-amadeus: I think it will all be clear after we design the final app's state. When we decide what and how `ab-tester` is going to analyze, it will be clear for us what and how we should implement.

PR #134, opened 2021-04-14
Fix / return winner even when there's no data

valentino-amadeus:
- To return the test's `winner` (winner over all workspaces) when there is no data to be analyzed.
- To replace `"master"` with `"Not yet decided"` as the `winner` value, when there is not enough data to decided which workspace is the winner.

ab-tester's `status` has been returning an array whose first entry is `{ Winner: <winner> }`, when at least one of the versions under test has had at least one session since the test's beginning. However, if there has been no sessions at all, `status` has not been returning such value. This is a problem, as it is an unexpected API contract that is hard to work with.

#### Types of changes


valentino-amadeus:
We'll stop returning the `winner` value for now, as its presence may break stuff.

